PR #2, opened 2019-01-09
 diffeq_adjoint is working

YingboMa:


src/Flux/layers.jl
```diff
@@ -38,11 +38,13 @@ function diffeq_adjoint(p,prob,f,df,t,args...;kwargs...)
...
-
 # Example: https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/blob/master/test/adjoint.jl
 # How to provide the cost function here? Is this fine?
 Flux.Tracker.@grad function diffeq_adjoint(p::TrackedVector,f,df,t,prob,args...;kwargs...)
   _prob = remake(prob,p=data(p))
   sol = solve(_prob,args...;kwargs...)
-  f(sol),Δ -> (Δ .* adjoint_sensitivities(sol,args...,df,t,kwargs...))
+  f(sol), Δ -> begin
+    grad = adjoint_sensitivities(sol,args...,df,t;sensealg=SensitivityAlg(autojacvec=false),kwargs...)
+    (Δ .* grad', ntuple(_->nothing, 4+length(args))...)
```
ChrisRackauckas: oh, we should tuple the fd one too for stability.

src/Flux/layers.jl
```diff
@@ -38,11 +38,13 @@ function diffeq_adjoint(p,prob,f,df,t,args...;kwargs...)
...
 diffeq_adjoint(p::TrackedVector,args...;kwargs...) = Flux.Tracker.track(diffeq_adjoint, p, args...; kwargs...)
-
 # Example: https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/blob/master/test/adjoint.jl
 # How to provide the cost function here? Is this fine?
 Flux.Tracker.@grad function diffeq_adjoint(p::TrackedVector,f,df,t,prob,args...;kwargs...)
   _prob = remake(prob,p=data(p))
   sol = solve(_prob,args...;kwargs...)
-  f(sol),Δ -> (Δ .* adjoint_sensitivities(sol,args...,df,t,kwargs...))
+  f(sol), Δ -> begin
+    grad = adjoint_sensitivities(sol,args...,df,t;sensealg=SensitivityAlg(autojacvec=false),kwargs...)
```
ChrisRackauckas: `SensitivityAlg(autojacvec=false)` why this choice? I think we should just make this a kwarg as well, and add it in front.

test/layers.jl
```diff
@@ -89,3 +89,27 @@ cb()
 Flux.train!(loss_fd2, params, data, opt, cb = cb)
 
 # Adjoint sensitivity
+ts = range(0.0,stop=10.0,step=0.1)
+p = param([2.2, 1.0, 2.0, 0.4])
+params = Flux.Params([p])
+loss_reduction′(out,u,p,t,i) = (out.=1.0.-u)
```
ChrisRackauckas: to make the test the same as the others, shouldn't it be `2*(u-1.0)`?

YingboMa:
#3

PR #5, opened 2019-01-21
Parameter munging utilities

MikeInnes:
Utilities for working with Flux models under the current version of the adjoint method. Only meant as a stopgap until I rework DiffEqSensitivities.jl.

```julia
julia> model = Chain(Dense(10, 5, relu), Dense(5, 2))
Chain(Dense(10, 5, NNlib.relu), Dense(5, 2))

julia> p = destructure(model)
Tracked 67-element Array{Float32,1}:
  0.35315588f0
  0.2662674f0
  0.042073943f0
  0.049423557f0
 -0.30862498f0
  0.35268134f0
  ⋮
  0.6537891f0
 -0.38841408f0
 -0.445121f0
  0.0f0
  0.0f0

julia> restructure(model, p)
Chain(Dense(10, 5, NNlib.relu), Dense(5, 2))
```

### DIFF
--- a/src/Flux/utils.jl
+++ b/src/Flux/utils.jl
@@ -1,0 +1,18 @@
+function destructure(m)
+  xs = []
+  mapleaves(m) do x
+    x isa TrackedArray && push!(xs, x)
+    return x
+  end
+  return vcat(vec.(xs)...)
+end
+
+function restructure(m, xs)
+  i = 0
+  mapleaves(m) do x
+    x isa TrackedArray || return x
+    x = reshape(xs[i.+(1:length(x))], size(x))
+    i += length(x)
+    return x
+  end
+end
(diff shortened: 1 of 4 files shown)


src/Flux/utils.jl
```diff
@@ -0,0 +1,18 @@
+function destructure(m)
+  xs = []
+  mapleaves(m) do x
+    x isa TrackedArray && push!(xs, x)
+    return x
+  end
+  return vcat(vec.(xs)...)
```
oxinabox: use `mapreduce(vec, vcat, xs)`
It is faster, due to the `reduce(vcat, xs)` hitting an optimized method

MikeInnes: It'd be best to time this in this case, since we're going to hit Flux's gradient definition anyway, and that potentially involves a lot of extra work.

In any case I'm not too worried about micro-optimising this, it's not going to be as fast as a real solution but just needs to work so we can prototype things.

PR #6, opened 2019-01-21
add neural ode and sde layers

ChrisRackauckas:
The idea is essentially

```julia
u0 = Float32[2.; 0.]
tspan = (0.0f0,25.0f0)

dudt = Chain(Dense(2,50,tanh),Dense(50,2))
p = Flux.data(destructure(dudt))
dudt_(du,u::TrackedArray,p,t) = du .= restructure(dudt,p)(u)
dudt_(du,u::AbstractArray,p,t) = du .= Flux.data(restructure(dudt,p)(u))
prob = ODEProblem(dudt_,u0,tspan,p)

x->diffeq_fd(p,reduction,2,prob,Tsit5(),u0=x,save_everystep=false)
```

but with a nice interface

### DIFF
--- a/src/Flux/neural_ode.jl
+++ b/src/Flux/neural_ode.jl
@@ -1,0 +1,15 @@
+neural_ode_reduction(sol) = Array(sol)
+function neural_ode(x,model,tspan,ad_func,args...;kwargs...)
+  p = Flux.data(destructure(model))
+  dudt_(du,u::TrackedArray,p,t) = du .= restructure(model,p)(u)
+  dudt_(du,u::AbstractArray,p,t) = du .= Flux.data(restructure(model,p)(u))
+  prob = ODEProblem(dudt_,x,tspan,p)
+
+  if ad_func === diffeq_adjoint
+    return ad_func(p,prob,args...;kwargs...)
+  elseif ad_func === diffeq_fd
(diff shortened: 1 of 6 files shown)


src/Flux/neural_ode.jl
```diff
@@ -0,0 +1,20 @@
...
+                                                         args...;kwargs...)
+function neural_ode(x,model,tspan,
+                    ad_func::Function,
+                    args...;kwargs...)
+  p = Flux.data(destructure(model))
+  dudt_(du,u::TrackedArray,p,t) = du .= restructure(model,p)(u)
+  dudt_(du,u::AbstractArray,p,t) = du .= Flux.data(restructure(model,p)(u))
+  prob = ODEProblem(dudt_,x,tspan,p)
+
+  if ad_func === diffeq_adjoint
```
oxinabox: This feels like it should be handled with a dispatch
more extensible that way.

Won't these methods fit onto `diffeq_adjoint` and `diffeq_fd` ?

Even if not 
a suitable proxy can be betteren.

`neural_ode_ad_func(::typeof(f), p, prob, args..., kwargs...)`

ChrisRackauckas: I'll merge for now, but a follow-up PR changing the API around is fine. I just want to get the internals solidified.

PR #8, opened 2019-01-22
make neural_ode work

MikeInnes:
minimal example:

```julia
using OrdinaryDiffEq, DiffEqFlux, Flux

dudt = Chain(Dense(2,50,tanh),Dense(50,2))

u0 = Float32[1.0,1.0]
tspan = (0.0,10.0)

n_ode = neural_ode(dudt,u0,tspan,Tsit5())

Flux.back!(sum(n_ode))

Tracker.grad(dudt[1].W)
```

### DIFF
--- a/src/Flux/layers.jl
+++ b/src/Flux/layers.jl
@@ -42,14 +42,14 @@ end
 
 diffeq_adjoint(p::TrackedVector,args...;kwargs...) = Flux.Tracker.track(diffeq_adjoint, p, args...; kwargs...)
 @grad function diffeq_adjoint(p::TrackedVector,prob,args...;
-                              u0=prob.u0,sensealg=SensitivityAlg(backsolve=true),kwargs...)
+                              u0=prob.u0,backsolve=false,kwargs...)
   _prob = remake(prob,u0=u0,p=Flux.data(p))
   sol = solve(_prob,args...;kwargs...)
   Array(sol), Δ -> begin
     Δ = Flux.data(Δ)
     ts = sol.t
     df(out, u, p, t, i) = @. out = - @view Δ[:, i]
-    grad = adjoint_sensitivities(sol,args...,df,ts;sensealg=sensealg,kwargs...)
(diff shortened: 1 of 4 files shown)


src/Flux/neural_de.jl
```diff
@@ -1,22 +1,11 @@
...
+function neural_ode(model,x,tspan,
                     args...;kwargs...)
-  p = Flux.data(destructure(model))
+  Tracker.istracked(x) && error("u0 is not currently differentiable.")
+  p = destructure(model)
   dudt_(du,u::TrackedArray,p,t) = du .= restructure(model,p)(u)
   dudt_(du,u::AbstractArray,p,t) = du .= Flux.data(restructure(model,p)(u))
   prob = ODEProblem(dudt_,x,tspan,p)
-
-  if ad_func === diffeq_adjoint
```
ChrisRackauckas: why remove the other options?

src/Flux/layers.jl
```diff
@@ -39,14 +41,15 @@ function diffeq_adjoint(p,prob,args...;u0=prob.u0,kwargs...)
 end
 
 diffeq_adjoint(p::TrackedVector,args...;kwargs...) = Flux.Tracker.track(diffeq_adjoint, p, args...; kwargs...)
-Flux.Tracker.@grad function diffeq_adjoint(p::TrackedVector,prob,args...;u0=prob.u0,kwargs...)
+@grad function diffeq_adjoint(p::TrackedVector,prob,args...;
+                              u0=prob.u0,sensealg=SensitivityAlg(backsolve=true),kwargs...)
```
ChrisRackauckas: we really shouldn't default to something that has a wide range of methods where it fails

src/Flux/neural_de.jl
MikeInnes: The other options don't really make sense here. Forward mode is never going to be useful for a neural ODE and for reverse mode will work without the destructure/restructure shenanigans (which add unnecessary overhead). Making AD work with the same signature is probably worthwhile but we can follow up with that.

PR #9, opened 2019-01-22
move the `f` outside of `diffeq_rd`

MikeInnes:


ChrisRackauckas:
Update the README as well

PR #13, opened 2019-01-24
Reverse-Mode Neural ODE

ChrisRackauckas:
This sets up an implementation of reverse-mode AD on neural ODEs. Everything uses TrackedArrays without mutation, which should be much better than Array{TrackedReal} due to the decreased tape size. This PR currently breaks the neural SDEs because those right now aren't compatible with TrackedArray, but that has a simple fix which I'll do to finish up the PR.


ChrisRackauckas:
Neural SDE now uses TrackedArray and passes.

PR #20, opened 2019-02-08
Simplify neural_ode_rd

ChrisRackauckas:


MikeInnes:
Working definition and test case:

```julia
function neural_ode_rd(model,x,tspan,
                       args...;kwargs...)
  dudt_(u,p,t) = model(u)
  prob = ODEProblem(dudt_,param(x),tspan)
  solve(prob, args...; kwargs...) |> Tracker.collect
end

dudt = Chain(Dense(2,50,tanh),Dense(50,2)) |> f64

u0 = [1.0,1.0]
tspan = (0.0,10.0)

n_ode = neural_ode_rd(dudt,u0,tspan,Tsit5())
```

You can avoid the `f64` if you also convert `u0` to f32; but I'll add some missing conversions to Flux so that this can't trip people up.

ChrisRackauckas:
Should we just do a bunch of conversions in our neural ODE functions to

MikeInnes:
I suspect it would be fiddly to put the right conversions in the right places, and the problem goes away once I tag https://github.com/FluxML/Flux.jl/pull/622.

PR #26, opened 2019-02-15
Simplify neural_ode_rd

MikeInnes:
Superceding #20


ChrisRackauckas:
Why did this fail? Is it waiting on a tag?

PR #42, opened 2019-04-10
Tests on GPU

jessebett:
I've left the reverse mode commented out. Also these tests will error but "pass" if the gradients error.


ChrisRackauckas:
I'll take a look at this when I get to a computer with a GPU tomorrow, thanks.

PR #52, opened 2019-06-03
Tracker -> Zygote

jessebett:
This is a WIP to move from Tracker to Zygote for AD.

Right now the strategy is to comment out everything except the adjoint solve. Delete all the `Tracker` things, and see what is failing where.

### DIFF
--- a/src/Flux/layers.jl
+++ b/src/Flux/layers.jl
@@ -1,49 +1,50 @@
-using Flux.Tracker: @grad
+#= using Flux.Tracker: @grad =#
+using Zygote: @adjoint
 using DiffEqSensitivity: adjoint_sensitivities_u0
 
 ## Reverse-Mode via Flux.jl
 
-function diffeq_rd(p,prob,args...;u0=prob.u0,kwargs...)
-  if typeof(u0) <: AbstractArray && !(typeof(u0) <: TrackedArray)
-    if DiffEqBase.isinplace(prob)
-      # use Array{TrackedReal} for mutation to work
-      # Recurse to all Array{TrackedArray}
-      _prob = remake(prob,u0=convert.(recursive_bottom_eltype(p),u0),p=p)
-    else
-      # use TrackedArray for efficiency of the tape
-      _prob = remake(prob,u0=convert(typeof(p),u0),p=p)
-    end
-  else # u0 is functional, ignore the change
-    _prob = remake(prob,u0=u0,p=p)
-  end
-  solve(_prob,args...;kwargs...)
-end
+#= function diffeq_rd(p,prob,args...;u0=prob.u0,kwargs...) =#
+#=   if typeof(u0) <: AbstractArray && !(typeof(u0) <: TrackedArray) =#
+#=     if DiffEqBase.isinplace(prob) =#
+#=       # use Array{TrackedReal} for mutation to work =#
+#=       # Recurse to all Array{TrackedArray} =#
+#=       _prob = remake(prob,u0=convert.(recursive_bottom_eltype(p),u0),p=p) =#
+#=     else =#
+#=       # use TrackedArray for efficiency of the tape =#
+#=       _prob = remake(prob,u0=convert(typeof(p),u0),p=p) =#
+#=     end =#
+#=   else # u0 is functional, ignore the change =#
+#=     _prob = remake(prob,u0=u0,p=p) =#
+#=   end =#
+#=   solve(_prob,args...;kwargs...) =#
+#= end =#
 
 ## Forward-Mode via ForwardDiff.jl
 
-function diffeq_fd(p,f,n,prob,args...;u0=prob.u0,kwargs...)
-  _prob = remake(prob,u0=convert.(eltype(p),u0),p=p)
-  f(solve(_prob,args...;kwargs...))
-end
-
-diffeq_fd(p::TrackedVector,args...;kwargs...) = Flux.Tracker.track(diffeq_fd, p, args...; kwargs...)
-Flux.Tracker.@grad function diffeq_fd(p::TrackedVector,f,n,prob,args...;u0=prob.u0,kwargs...)
-  _f = function (p)
-    _prob = remake(prob,u0=convert.(eltype(p),u0),p=p)
-    f(solve(_prob,args...;kwargs...))
-  end
-  _p = Flux.data(p)
-  if n === nothing
-    result = DiffResults.GradientResult(_p)
-    ForwardDiff.gradient!(result, _f, _p)
-    DiffResults.value(result),Δ -> (Δ .* DiffResults.gradient(result), ntuple(_->nothing, 3+length(args))...)
-  else
-    y = zeros(n)
-    result = DiffResults.JacobianResult(y,_p)
-    ForwardDiff.jacobian!(result, _f, _p)
-    DiffResults.value(result),Δ -> (DiffResults.jacobian(result)' * Δ, ntuple(_->nothing, 3+length(args))...)
-  end
-end
+#= function diffeq_fd(p,f,n,prob,args...;u0=prob.u0,kwargs...) =#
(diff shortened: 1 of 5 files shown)


jessebett:
@ChrisRackauckas 

```
using Flux, DiffEqFlux, OrdinaryDiffEq
using Zygote

u0 = Float32[2.; 0.]
datasize = 30
tspan = (0.0f0,1.5f0)

function trueODEfunc(du,u,p,t)
    true_A = [-0.1 2.0; -2.0 -0.1]
    du .= ((u.^3)'true_A)'
end
t = range(tspan[1],tspan[2],length=datasize)
prob = ODEProblem(trueODEfunc,u0,tspan)
ode_data = Array(solve(prob,Tsit5(),saveat=t))

dudt = Chain(x -> x.^3,
             Dense(2,50,tanh),
             Dense(50,2))
ps = Flux.params(dudt)

p = DiffEqFlux.destructure(dudt)
dudt_(u,p,t) = DiffEqFlux.restructure(dudt,p)(u)
prob = ODEProblem(dudt_,u0,tspan,p)

diffeq_adjoint(p,prob,Tsit5(), u0=u0) #A

forward(u->diffeq_adjoint(p,prob,Tsit5(), u0=u),u0) #B
```

The forward pass in #A works, but not when called like in #B. It fails with error `ERROR: MethodError: no method matching isinplace(::Tsit5)` Thoughts on where this could be coming from?

src/Flux/layers.jl
```diff
@@ -54,16 +55,16 @@ function diffeq_adjoint(p,prob,args...;u0=prob.u0,kwargs...)
   adapt(T, solve(_prob,args...;kwargs...))
 end
 
-diffeq_adjoint(p::TrackedVector,prob,args...;u0=prob.u0,kwargs...) =
-  Flux.Tracker.track(diffeq_adjoint, p, u0, prob, args...; kwargs...)
+#= diffeq_adjoint(p::TrackedVector,prob,args...;u0=prob.u0,kwargs...) = =#
+#=   Flux.Tracker.track(diffeq_adjoint, p, u0, prob, args...; kwargs...) =#
 
-@grad function diffeq_adjoint(p,u0,prob,args...;backsolve=true,
+@adjoint function diffeq_adjoint(p,u0,prob,args...;backsolve=true,
```
ChrisRackauckas: it's this right here

ChrisRackauckas:
It thinks the `alg` is the `prob` because of that adjoint statement. Change this to `forward(u->diffeq_adjoint(p,u,prob,Tsit5()),u0) #B`

src/Flux/layers.jl
jessebett: Why is this not an issue with Tracker?

```
diffeq_adjoint(p::TrackedVector,prob,args...;u0=prob.u0,kwargs...) =
  Flux.Tracker.track(diffeq_adjoint, p, u0, prob, args...; kwargs...)


@grad function diffeq_adjoint(p,u0,prob,args...;backsolve=true,
                              save_start=true,
                              sensealg=SensitivityAlg(quad=false,backsolve=backsolve),
                              kwargs...)
```

https://github.com/JuliaDiffEq/DiffEqFlux.jl/blob/a8105ee042d1f80538d428f1e50acf421f572cde/src/Flux/layers.jl#L57-L63

Why is this not an issue with Tracker?

jessebett: Ah okay. Do Zygote adjoints not need separate function definitions (i.e. without and with `@adjoint`)? If they do then which argument order is correct? Your suggested change has `p,u0,prob...` but the Tracker (non-`@grad`) has `u0` a keyword argument whose default value is supplied by the `prob`. Which should I go with?

ChrisRackauckas: I guess add the argument

jessebett:
@MikeInnes the argument order changes @ChrisRackauckas pointed out was blocking more Zygote errors. 

I'll leave some comments on `test/simple-neurode.jl` which is the small script I'm running to test Zygote integration.

test/simple-neurode.jl
```diff
@@ -0,0 +1,48 @@
...
+#= ps = Flux.params(dudt) =#
+
+p = DiffEqFlux.destructure(dudt)
+ps = Flux.params(p)
+dudt_(u,p,t) = DiffEqFlux.restructure(dudt,p)(u)
+prob = ODEProblem(dudt_,u0,tspan,p)
+
+diffeq_adjoint(p,u0,prob,Tsit5())
+
+forward(u->diffeq_adjoint(p,u,prob,Tsit5()),u0)
```
jessebett: @MikeInnes evaluates, but evaluating the pullback gives the variable not defined error.

```
f_, back = forward(u->diffeq_adjoint(p,u,prob,Tsit5()),u0)
back(f_) #ERROR: Variable %46 in block 8 is not defined.
```

Also this is true for `p` as well:
`f_, back = forward(p_->diffeq_adjoint(p_,u0,prob,Tsit5()),p)`

jessebett: @MikeInnes correction, with updates to Zygote#master this now gives `ERROR: Can't differentiate foreigncall expression`

yuehhua:
Does this need any help?

jessebett:
Hi @yuehhua since this draft PR was made there has been a lot of progress so that Flux uses Zygote by default, which is now true of Flux#master. I don't remember the specific zygote reasons that caused the code in `test/simple-neurode.jl` to fail. If you're up for it you could try to run that code against `Flux#master` and report what issues come up?

There will definitely be some things that need to change in `DiffEqFlux`. In particular we dispatch the dynamics differently depending on whether they're tracked or not, which will not be an issue with Zygote.

yuehhua:
@jessebett Sure! I will check the code here and gradually update to Zygote.

yuehhua:
While I test on

```
grads = Zygote.gradient(loss_rd, params)
```

and I got this:

```
Got exception outside of a @test
  LoadError: Need an adjoint for constructor DiffEqBase.ODESolution{Float64,2,Array{Array{Float64,1},1},Nothing,Nothing,Array{Float64,1},Array{Array{Array{Float64,1},1},1},DiffEqBase.ODEProblem{Array{Float64,1},Tuple{Float64,Float64},true,Array{Float64,1},DiffEqBase.ODEFunction{true,typeof(Main.##486.lotka_volterra),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{,Tuple{}}},DiffEqBase.StandardODEProblem},OrdinaryDiffEq.Tsit5,OrdinaryDiffEq.InterpolationData{DiffEqBase.ODEFunction{true,typeof(Main.##486.lotka_volterra),LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},Array{Array{Float64,1},1},Array{Float64,1},Array{Array{Array{Float64,1},1},1},OrdinaryDiffEq.Tsit5Cache{Array{Float64,1},Array{Float64,1},Array{Float64,1},OrdinaryDiffEq.Tsit5ConstantCache{Float64,Float64}}},DiffEqBase.DEStats}. Gradient is of type Array{Float64,2}
```

ChrisRackauckas:
@MikeInnes is there a way to make it not try to take an adjoint of the DESolution constructors? I thought that would be something it would know automatically since it's just putting values into a holder.

MikeInnes:
If it was really just a container, that would be true, but AFAIK DESolution gets used as an array which means the gradient will be an array. Zygote can't automatically figure out how to map the array to gradients of the DESolution components, but it should be pretty easy to write down something that handles that.

ChrisRackauckas:
Supersceded by https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/96

PR #60, opened 2019-07-17
add gpu tests

ChrisRackauckas:


test/runtests.jl
```diff
@@ -1,9 +1,22 @@
...
 
-@safetestset "Utils Tests" begin include("utils.jl") end
-@safetestset "Layers Tests" begin include("layers.jl") end
-@safetestset "Layers SDE" begin include("layers_sde.jl") end
-@safetestset "Layers DDE" begin include("layers_dde.jl") end
-@safetestset "odenet" begin include("odenet.jl") end
-@safetestset "Neural DE Tests" begin include("neural_de.jl") end
-@safetestset "Partial Neural Tests" begin include("partial_neural.jl") end
+const GROUP = get(ENV, "GROUP", "All")
+const is_APPVEYOR = ( Sys.iswindows() && haskey(ENV,"APPVEYOR") )
```
oxinabox:
```suggestion
const is_CI = haskey(ENV,"CI")
```

ChrisRackauckas: I want to keep `is_APPVEYOR` because I use it all of the time to disable tests from AppVeyor, since my AppVeyor account is usually 2 days behind already.

oxinabox: I didn't delete it?

ChrisRackauckas: oh, I read the suggestion incorrectly.

ChrisRackauckas: it'll get added when it comes into play. Thanks, I didn't know that env variable existed.

PR #66, opened 2019-07-20
Respect users' saving option

### LINKED ISSUE #64: save_start=false argument ignored for neural_ode solve
Tested here: https://github.com/jessebett/DiffEqFlux.jl/blob/master/test/solver_options.jl

If you call `neural_ode` with `save_start=false` it still returns the start. I believe this needed to be overridden internally to get adjoint to work, but what is returned should respect this argument.

It is very common to run these solves only requiring the output to be saved. It would be nice if there was an option like `save_only` in this case would be the end of the timespan.

YingboMa:
Fix #64


ChrisRackauckas:
💇‍♂ 😹 🥇 🖐

PR #73, opened 2019-07-25
NeuralODE struct interface

jessebett:
Exporting a struct for `NeuralODE` that dispatches to `neural_ode` solve and whose `params` expose the underlying `dudt` to Flux makes the API nicer.

### DIFF
--- a/src/Flux/neural_de.jl
+++ b/src/Flux/neural_de.jl
@@ -1,24 +1,10 @@
-struct NeuralODE
-    model
-    tspan
-    solver
-    kwargs
-end
-Flux.@treelike NeuralODE
-
-function Flux.params(n::NeuralODE) = params(n.model)
-
-function (n::NeuralODE)(x)
-    return neural_ode(n.model,x,n.tspan,n.solver, n.kwargs...)
-end
-
-function neural_ode(model,x,tspan,
+function neural_ode(model,x,tspan, solver,
                     args...;kwargs...)
   p = destructure(model)
   dudt_(u::TrackedArray,p,t) = restructure(model,p)(u)
   dudt_(u::AbstractArray,p,t) = Flux.data(restructure(model,p)(u))
   prob = ODEProblem(dudt_,x,tspan,p)
-  return diffeq_adjoint(p,prob,args...;u0=x,kwargs...)
+  return diffeq_adjoint(p,prob,solver,args...;u0=x,kwargs...)
 end
 
 function neural_ode_rd(model,x,tspan,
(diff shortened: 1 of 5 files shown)


src/Flux/neural_de.jl
```diff
@@ -1,10 +1,10 @@
-function neural_ode(model,x,tspan,
+function neural_ode(model,x,tspan, solver,
```
jessebett: I pulled `solver` out of `args` so it is an explicit field in the struct. I don't know if this is a good idea... @ChrisRackauckas 

Also what are some other examples of `args...` besides  the solver choice?

src/Flux/neural_de.jl
```diff
@@ -24,3 +24,24 @@ function neural_dmsde(model,x,mp,tspan,
...
+struct NeuralODE
+    model
+    tspan
+    solver
+    args
+    kwargs
+end
+
+# Optional args and kwargs
+NeuralODE(model,tspan,solver,kwargs)= NeuralODE(model,tspan,solver,(),kwargs)
```
jessebett: This is related to the `args...` question. Is there a better way to allow both `args` and `kwargs` to be optional?

test/solver_options.jl
```diff
@@ -0,0 +1,15 @@
+using OrdinaryDiffEq, StochasticDiffEq, Flux, DiffEqFlux, Test
+
+xs = Float32.(hcat([0.; 0.], [1.; 0.], [2.; 0.]))
+tspan = (0.0f0,25.0f0)
+dudt = Chain(Dense(2,50,tanh),Dense(50,2))
+
+@testset "only end" begin
```
jessebett: These tests need to be expanded to include other solver options. Especially the various ones which control the timepoints returned. I've noticed some combination of `save_*` break when trying to only return the final time step.

test/neural_de_gpu.jl
```diff
@@ -146,3 +146,22 @@ end;
...
+    nn = Chain(
+               Dense(20,10,tanh),
+               Dense(10,10,tanh),
+               Dense(10,20,tanh)
+              ) |>gpu
+    fc = Chain(
+               Dense(20,10)
+              )|>gpu
+    nn_ode(x) = neural_ode(nn,x,gpu((0.f0,1.f0)), Tsit5(),save_everystep=false,reltol=1e-3,abstol=1e-3, save_start=false)
+    @test_broken typeof(nn(down(x))) == typeof(nn_ode(down(x)))
```
jessebett: I don't think this is still broken. This test needs to include a `allowscalar(false)` I will update this test tomorrow.

src/Flux/neural_de.jl
ChrisRackauckas: Just use the solver one

src/Flux/neural_de.jl
ChrisRackauckas:
```suggestion
NeuralODE(model,tspan,solver,args...;kwargs...)= NeuralODE(model,tspan,solver,args,kwargs)
```

ChrisRackauckas: The suggestion I did will naturally allow passing both and pull the splats into the appropriate fields

test/solver_options.jl
ChrisRackauckas: interesting, yes please share

test/neural_de_gpu.jl
ChrisRackauckas: where is `down(x)` from?

test/solver_options.jl
ChrisRackauckas: these should all be fine nowadays.

ChrisRackauckas:
Supersceded by https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/109

PR #76, opened 2019-09-12
Replace Flux with Tracker

ChrisRackauckas:


ChrisRackauckas:
Flux is required because it needs to define `children` for Dense.

PR #86, opened 2019-11-18
Add Plots as a test dependency

dhairyagandhi96:
The CI fails without it, also there is a separate issue with checking the gradients at various places.


ChrisRackauckas:
It needs to go the other way. That was a mistake: https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/87

PR #96, opened 2019-12-14
Zygote Compatibility

ChrisRackauckas:
Requires https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/400 . Everything works except for the neural networks... this is because destructure drops gradients. The issue is because `push!` is not differentiable. @mikeinnes do you have a quick fix to finish this?


ChrisRackauckas:
Basically, how to make this differentiable?

https://github.com/JuliaDiffEq/DiffEqFlux.jl/blob/master/src/Flux/utils.jl#L1-L8

ChrisRackauckas:
Requires https://github.com/FluxML/Zygote.jl/issues/427 for tests to pass.

src/Flux/neural_de.jl
```diff
@@ -21,12 +21,12 @@ kwargs key word arguments passed to ODESolve; accepts an additional key
...
 """
-function neural_ode(model,x,tspan,
-                    args...;kwargs...)
+function neural_ode(model,x,tspan,args...;kwargs...)
   p = destructure(model)
-  dudt_(u::TrackedArray,p,t) = restructure(model,p)(u)
-  dudt_(u::AbstractArray,p,t) = Tracker.data(restructure(model,p)(u))
-  prob = ODEProblem(dudt_,x,tspan,p)
+  _model = restructure(model,p)
+  dudt_(u::TrackedArray,p,t) = _model(u)
```
jessebett: @ChrisRackauckas Zygote compatibility progress! Awesome! these lines will go away also, right? No need to dispatch separately on Tracked Arrays now.

src/Flux/utils.jl
```diff
@@ -1,16 +1,25 @@
 function destructure(m)
```
jessebett: @MikeInnes can we make this a utility in Flux instead of DiffEqFlux? 
@ChrisRackauckas idk if you remember but on my fork of DiffEqFlux I have destructure/restructure returning slightly different objects and I think we might prefer them. Essentially have `destructure` return the vector and an anonymous function which `restructure`s. Removes the need for a separate `restructure` function and also that dependence on `m`, which is a little confusing, since it only uses `m` for its structure, not its current parameter values. What do you think?

dhairyagandhi96: I had a similar plan of action in mind, with the params wrapped in a `Buffer`. The anonymous function sounds about right too. Can we `loadparams!` on `m` by any chance? That would be cleaner, but might need some thought.

src/Flux/neural_de.jl
ChrisRackauckas: For now we need to keep Tracker compatibility for the adjoint pass.

dhairyagandhi96:
The context approach seems a bit fragile to me. Would it be better to just hide the `push!`? I have a prototype here:
https://github.com/dhairyagandhi96/Zygote.jl/blob/37185d5257c563dce3623f743aea4353c56ad6bc/src/compiler/interface.jl#L63

ChrisRackauckas:
>The context approach seems a bit fragile to me. Would it be better to just hide the push!? I have a prototype here:

But does that actually work? This is the first thing that I actually got working, so it's at least less fragile than Zygote :P.

dhairyagandhi96:
Fwiw, I used this to get things working for me. It does need a function in Flux to forward the `Params` construction, but that's about all

ChrisRackauckas:
I'll get this green and merge, and then we'll continue before any tagging. `diffeq_adjoint` seems to be fine: it's directly editing the context but it's doing so in a way that's not too unsafe. I'm not as worried there. But `diffeq_rd` has issues. It's supposed to be making the Zygote method fallback to Tracker, but that's hard to do because... implicit parameters are hard to make TrackedArrays. Do you have a potential solution there?

dhairyagandhi96:
Oh, interesting. Will get check that out.

ChrisRackauckas:
Okay, tests are green. This still isn't great, but it's something to build off of. @dhairyagandhi96 let's keep improving before the tag.

homocomputeris:
Does "everything except NNs" mean that it is not compatible with NeuralNetDiffEq?

```
add Flux@0.10
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package DiffEqBase [2b5f629d]:
 DiffEqBase [2b5f629d] log:
 ├─possible versions are: [3.13.2-3.13.3, 4.0.0-4.0.1, 4.1.0, 4.2.0, 4.3.0-4.3.1, 4.4.0, 4.5.0, 4.6.0, 4.7.0, 4.8.0, 4.9.0, 4.10.0-4.10.1, 4.11.0-4.11.1, 4.12.0, 4.13.0, 4.14.0-4.14.1, 4.15.0, 4.16.0, 4.17.0, 4.18.0, 4.19.0, 4.20.0-4.20.3, 4.21.0, 4.21.2-4.21.3, 4.22.0-4.22.2, 4.23.0, 4.23.2-4.23.4, 4.24.0-4.24.3, 4.25.0-4.25.1, 4.26.0-4.26.3, 4.27.0-4.27.1, 4.28.0-4.28.1, 4.29.0-4.29.2, 4.30.0-4.30.2, 4.31.0-4.31.2, 4.32.0, 5.0.0-5.0.1, 5.1.0, 5.2.0-5.2.3, 5.3.0-5.3.2, 5.4.0-5.4.1, 5.5.0-5.5.2, 5.6.0-5.6.4, 5.7.0, 5.8.0-5.8.1, 5.9.0, 5.10.0-5.10.3, 5.11.0-5.11.1, 5.12.0, 5.13.0, 5.14.0-5.14.2, 5.15.0, 5.16.0-5.16.5, 5.17.0-5.17.1, 5.18.0, 5.19.0, 5.20.0-5.20.1, 6.0.0, 6.1.0, 6.2.0-6.2.4, 6.3.0-6.3.6, 6.4.0-6.4.2, 6.5.0-6.5.1, 6.6.0, 6.7.0, 6.8.0, 6.9.0-6.9.4, 6.10.0-6.10.2, 6.11.0, 6.12.0-6.12.1] or uninstalled
 ├─restricted by compatibility requirements with NeuralNetDiffEq [8faf48c0] to versions: [5.0.0-5.0.1, 5.1.0, 5.2.0-5.2.3, 5.3.0-5.3.2, 5.4.0-5.4.1, 5.5.0-5.5.2, 5.6.0-5.6.4, 5.7.0, 5.8.0-5.8.1, 5.9.0, 5.10.0-5.10.3, 5.11.0-5.11.1, 5.12.0, 5.13.0, 5.14.0-5.14.2, 5.15.0, 5.16.0-5.16.5, 5.17.0-5.17.1, 5.18.0, 5.19.0, 5.20.0-5.20.1, 6.0.0, 6.1.0, 6.2.0-6.2.4, 6.3.0-6.3.6, 6.4.0-6.4.2, 6.5.0-6.5.1, 6.6.0, 6.7.0, 6.8.0, 6.9.0-6.9.4, 6.10.0-6.10.2, 6.11.0, 6.12.0-6.12.1]
 │ └─NeuralNetDiffEq [8faf48c0] log:
 │   ├─possible versions are: [0.0.1, 0.1.0, 1.0.0-1.0.2, 1.1.0] or uninstalled
 │   └─restricted to versions 1.0.1 by an explicit requirement, leaving only versions 1.0.1
 ├─restricted by compatibility requirements with DifferentialEquations [0c46a032] to versions: [6.6.0, 6.7.0, 6.8.0, 6.9.0-6.9.4, 6.10.0-6.10.2, 6.11.0, 6.12.0-6.12.1]
 │ └─DifferentialEquations [0c46a032] log:
 │   ├─possible versions are: [5.0.0, 5.1.0, 5.2.0-5.2.1, 5.3.0-5.3.1, 6.0.0, 6.1.0, 6.2.0, 6.3.0, 6.4.0, 6.6.0, 6.7.0, 6.8.0, 6.9.0, 6.10.0] or uninstalled
 │   └─restricted to versions 6.9.0 by an explicit requirement, leaving only versions 6.9.0
 └─restricted by compatibility requirements with DiffEqFlux [aae7a2af] to versions: [3.13.2-3.13.3, 4.0.0-4.0.1, 4.1.0, 4.2.0, 4.3.0-4.3.1, 4.4.0, 4.5.0, 4.6.0, 4.7.0, 4.8.0, 4.9.0, 4.10.0-4.10.1, 4.11.0-4.11.1, 4.12.0, 4.13.0, 4.14.0-4.14.1, 4.15.0, 4.16.0, 4.17.0, 4.18.0, 4.19.0, 4.20.0-4.20.3, 4.21.0, 4.21.2-4.21.3, 4.22.0-4.22.2, 4.23.0, 4.23.2-4.23.4, 4.24.0-4.24.3, 4.25.0-4.25.1, 4.26.0-4.26.3, 4.27.0-4.27.1, 4.28.0-4.28.1, 4.29.0-4.29.2, 4.30.0-4.30.2, 4.31.0-4.31.2, 4.32.0, 5.0.0-5.0.1, 5.1.0, 5.2.0-5.2.3, 5.3.0-5.3.2, 5.4.0-5.4.1, 5.5.0-5.5.2, 5.6.0-5.6.4, 5.7.0, 5.8.0-5.8.1, 5.9.0, 5.10.0-5.10.3, 5.11.0-5.11.1, 5.12.0, 5.13.0, 5.14.0-5.14.2, 5.15.0, 5.16.0-5.16.5, 5.17.0-5.17.1, 5.18.0, 5.19.0, 5.20.0-5.20.1] — no versions left
   └─DiffEqFlux [aae7a2af] log:
     ├─possible versions are: [0.0.1, 0.1.0, 0.2.0, 0.3.0, 0.4.0, 0.5.0-0.5.2, 0.6.0-0.6.1, 0.7.0, 0.8.0-0.8.1, 0.9.0, 0.10.0-0.10.2] or uninstalled
     ├─restricted by compatibility requirements with NeuralNetDiffEq [8faf48c0] to versions: [0.0.1, 0.1.0, 0.2.0, 0.3.0, 0.4.0, 0.5.0-0.5.2, 0.6.0-0.6.1, 0.7.0, 0.8.0-0.8.1, 0.9.0, 0.10.0-0.10.2]
     │ └─NeuralNetDiffEq [8faf48c0] log: see above
     └─restricted by compatibility requirements with Flux [587475ba] to versions: [0.0.1, 0.1.0, 0.2.0, 0.3.0, 0.4.0] or uninstalled, leaving only versions: [0.0.1, 0.1.0, 0.2.0, 0.3.0, 0.4.0]
       └─Flux [587475ba] log:
         ├─possible versions are: [0.4.1, 0.5.0-0.5.4, 0.6.0-0.6.10, 0.7.0-0.7.3, 0.8.0-0.8.3, 0.9.0, 0.10.0] or uninstalled
         └─restricted to versions 0.10 by an explicit requirement, leaving only versions 0.10.0

```

ChrisRackauckas:
You need a specific branch of Flux right now. https://github.com/FluxML/Flux.jl/pull/986#issuecomment-573004184 when that merges we will tag and update NeuralNetDiffEq

PR #97, opened 2019-12-18
[WIP] Refactoring destructure

dhairyagandhi96:
Needs a fix on Zygote to get accumulation over `Params` objects.

Ref #96 

This replaces the `destructure` implementation with one that uses `Zygote.Buffer` to get around array mutation. It currently adds the Buffer to the list of params, but that can be ignored.

The values of the grads shouldn't be zero though, so will need to fix that.

cc @ChrisRackauckas


dhairyagandhi96:
Also ref https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/96#discussion_r358378616

ChrisRackauckas:
this'll do.

PR #123, opened 2020-01-29
[WIP] Add sciml_train! function

abhigupta768:
Adds `sciml_train!` function

### DIFF
--- a/src/train.jl
+++ b/src/train.jl
@@ -1,0 +1,23 @@
+function cb(opt_state:: Optim.OptimizationState)
+    cur_pred = collect(predict_adjoint(opt_state.metadata["x"]))
+    n = size(training_data, 1)
+    pl = scatter(1:n,training_data[:,10],label="data", legend =:bottomright,title="Spatial Plot at t=$(saveat[10])")
+    scatter!(pl,1:n,cur_pred[:,10],label="prediction")
+    pl2 = scatter(saveat,training_data[N÷2,:],label="data", legend =:bottomright, title="Timeseries Plot at Middle X")
+    scatter!(pl2,saveat,cur_pred[N÷2,:],label="prediction")
+    display(plot(pl, pl2, size=(600, 300)))
+    display(opt_state.value)
+    false
+end
+
+cb(trace::Optim.OptimizationTrace) = cb(last(trace))
+
+function sciml_train!(loss_f, params, data, opt::Optim.AbstractOptimizer;cb = cb)
+    result =  optimize(loss_f, params, opt, Optim.Options(extended_trace=true,callback = cb))
+    return result
+end
+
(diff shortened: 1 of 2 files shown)


src/train.jl
```diff
@@ -0,0 +1,23 @@
+function cb(opt_state:: Optim.OptimizationState)
+    cur_pred = collect(predict_adjoint(opt_state.metadata["x"]))
```
ChrisRackauckas: users need to write the callback

src/train.jl
```diff
@@ -0,0 +1,23 @@
...
+    pl2 = scatter(saveat,training_data[N÷2,:],label="data", legend =:bottomright, title="Timeseries Plot at Middle X")
+    scatter!(pl2,saveat,cur_pred[N÷2,:],label="prediction")
+    display(plot(pl, pl2, size=(600, 300)))
+    display(opt_state.value)
+    false
+end
+
+cb(trace::Optim.OptimizationTrace) = cb(last(trace))
+
+function sciml_train!(loss_f, params, data, opt::Optim.AbstractOptimizer;cb = cb)
```
ChrisRackauckas: Does this actually work? Needs tests

src/train.jl
```diff
@@ -0,0 +1,23 @@
...
+    result =  optimize(loss_f, params, opt, Optim.Options(extended_trace=true,callback = cb))
+    return result
+end
+
+function sciml_train!(loss_f, params, data, opt::Union{Flux.Descent, Flux.Momentum, Flux.Nesterov, Flux.RMSProp, Flux.ADAM, Flux.RADAM, Flux.AdaMax, Flux.ADAGrad, Flux.ADADelta, Flux.AMSGrad, Flux.NADAM, Flux.ADAMW, Flux.Optimiser, Flux.InvDecay, Flux.ExpDecay, Flux.WeightDecay};cb = ())
```
ChrisRackauckas: is there no abstract type for Flux?

src/train.jl
abhigupta768: I see, so should it default to `()`?

src/train.jl
abhigupta768: Can you suggest a test I should use?

src/train.jl
abhigupta768: Not as far as I could see, https://github.com/FluxML/Flux.jl/blob/master/src/optimise/optimisers.jl

src/train.jl
Vaibhavdixit02: https://github.com/JuliaDiffEq/DiffEqFlux.jl#universal-differential-equations @ChrisRackauckas the example here with Flux and Optim should be a good MWE to build on?

src/train.jl
ChrisRackauckas: yes

src/train.jl
ChrisRackauckas: Send a PR to Flux to fix that.

src/train.jl
ChrisRackauckas: Yes exactly. It should be pretty clear what the interface is to most users of the package. I think the confusion might be just from not using it. Give the tutorials and everything a run, tweak them around, play with them, and see the use case is like. Then the training loop changes explanation will make sense, and you'll see what needs to be done with this PR. Right now this PR doesn't make any sense.

ChrisRackauckas:
Supersceded by https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/125

PR #136, opened 2020-02-04
add some abstract types

ChrisRackauckas:


cems2:
Add an Abstract type for NeuralODE.  I think it should be ::Function at a minimum since instances quack like functions and it's predecessor neural_ode() had an abstract ::Function type

ChrisRackauckas:
Already did.

PR #139, opened 2020-02-05
stopping criterion on current loss

ric-cioffi:
We can make this directly depend on the callback function (which is consistent with Optim).
For example a reasonable callback function would be
```
function cb(θ, l)
  println("Current loss: ", l)
  return l < 1e-4
end
```


ChrisRackauckas:
Yeah, I was wondering if we should do this, but in a safer way. I will finish this up.

PR #147, opened 2020-02-09
Add table of contents to README

ali-ramadhan:
README is a little hard to navigate right now so I added a table of contents to it until it's moved to documentation.

Following section names might have to be simplified to get usable anchor tag names.
* Neural Differential Equations for Non-ODEs: Neural SDEs, Neural DDEs, etc.
* Neural Jump Diffusions (Neural Jump SDE) and Neural Partial Differential Equations (Neural PDEs)

Suggestion for new section names:
* Neural Differential Equations for Non-ODEs
* Neural Jump Diffusions and Neural Partial Differential Equations


ChrisRackauckas:
Yes thanks .We're going to move to proper docs soon

PR #151, opened 2020-02-10
[WIP] Add box constrained optimization dispatch in sciml_train

abhigupta768:
@ChrisRackauckas @Vaibhavdixit02

### DIFF
--- a/src/train.jl
+++ b/src/train.jl
@@ -89,3 +89,26 @@ function sciml_train(loss, θ, opt::Optim.AbstractOptimizer;
            Optim.Options(extended_trace=true,callback = _cb,
                          f_calls_limit = maxiters))
 end
+
+function sciml_train(loss, lower_bounds, upper_bounds, θ, opt::Optim.AbstractConstrainedOptimizer;
+                      cb = (args...) -> (false), maxiters = 0)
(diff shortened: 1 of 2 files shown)


src/train.jl
```diff
@@ -89,3 +89,26 @@ function sciml_train(loss, θ, opt::Optim.AbstractOptimizer;
            Optim.Options(extended_trace=true,callback = _cb,
                          f_calls_limit = maxiters))
 end
+
+function sciml_train(loss, lower_bounds, upper_bounds, θ, opt::Optim.AbstractConstrainedOptimizer;
```
ChrisRackauckas: move lower and upper bounds to required kwargs. I never thought this was a good idea by Optim

Vaibhavdixit02:
This doesn't cover univariate problems https://github.com/JuliaNLSolvers/Optim.jl/blob/master/docs/src/user/minimization.md#minimizing-a-univariate-function-on-a-bounded-interval though. @ChrisRackauckas should we add that as well? I am not sure wrt the use cases.

ChrisRackauckas:
I don't think we have a need for it 🤷‍♂ , but open a low priority issue.

PR #152, opened 2020-02-11
Utilize view strided fixes

ChrisRackauckas:
Should get merged after https://github.com/JuliaLang/julia/pull/34729 with an `if` for the version number.

@mbauman thanks!


mbauman:
Unfortunately it's not easily monkey-patchable as it depends on slightly changing the `StridedArray` union definition.

ChrisRackauckas:
That's okay, we can just `if VERSION` kind of thing to upgrade when it's available. It'll give me a reason to update 👍

PR #154, opened 2020-02-12
Minibatching example

### LINKED ISSUE #133: Add an example with minibatching for sciml_train
Currently none of the README examples showcase how to minibatch. It would be a good thing to teach users.

ali-ramadhan:
Not sure if this is what you had in mind @ChrisRackauckas by moving the batching/data selection into the loss function, but I'm happy to revise and work on this to turn it into a useful example (I think it's far from useful right now).

Well the minibatching isn't really done cleanly. Instead of comparing the entire ODE time series in the loss function, it compares next state vs. predicted state and cycles through all the available data. This is the approach I've been taking when my data is coming from multiple simulations.

Added it as `minibatching.jl` so it's easier to look at the code but I can easily move it to the README with all the other examples.

Resolves #133

### DIFF
--- a/examples/minibatching.jl
+++ b/examples/minibatching.jl
@@ -1,0 +1,55 @@
+using DiffEqFlux, OrdinaryDiffEq, Flux, Optim, Plots
+
+u0 = Float32[2.; 0.]
+datasize = 30
+tspan = (0.0f0,1.5f0)
+
+function trueODEfunc(du,u,p,t)
+    true_A = [-0.1 2.0; -2.0 -0.1]
+    du .= ((u.^3)'true_A)'
+end
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Yes, this is kind of what I had in mind. As an example I was thinking of having a big time series data and doing things like randomly sampling or iteratively subsetting it for the fit, but that's just a different example. In general, this is how I saw it working. Any issues you see with this?

@colinwarner @jessebett

ali-ramadhan:
I believe this PR/example is unnecessary now? Close along with #133?

ChrisRackauckas:
This exact one, yes. But we still have no documentation or example of the data iterator system, so it would be good to show that somewhere.

PR #159, opened 2020-02-17
[WIP] Add neural DAE support

abhigupta768:
@ChrisRackauckas @Vaibhavdixit02

### DIFF
--- a/src/neural_de.jl
+++ b/src/neural_de.jl
@@ -168,9 +168,11 @@ function (n::NeuralCDDE)(x,p=n.p)
     concrete_solve(prob,n.solver,x,p,n.args...;sensealg=TrackerAdjoint(),n.kwargs...)
 end
 
-struct NeuralDAE{P,M,RE,T,S,DV,A,K} <: NeuralDELayer
+struct NeuralDAE{P,M,M2,D,RE,T,S,DV,A,K} <: NeuralDELayer
     model::M
+    constraints_model::M2
     p::P
+    du0::D
     re::RE
     tspan::T
     solver::S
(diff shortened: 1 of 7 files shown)


src/neural_de.jl
```diff
@@ -167,3 +167,29 @@ function (n::NeuralCDDE)(x,p=n.p)
     prob = DDEProblem{false}(dudt_,x,n.hist,n.tspan,p,constant_lags = n.lags)
     concrete_solve(prob,n.solver,x,p,n.args...;sensealg=TrackerAdjoint(),n.kwargs...)
 end
+
+struct NeuralDAE{P,M,RE,T,S,DV,A,K} <: NeuralDELayer
```
Vaibhavdixit02: This does not cover the mass matrix form. That should be added as per https://docs.juliadiffeq.org/latest/tutorials/advanced_ode_example/#Handling-Mass-Matrices-1 and the discussion in the slack thread.

Vaibhavdixit02:
This would need a test to build on, @ChrisRackauckas can you point me to an example how that should like?

ChrisRackauckas:
A mass matrix one would be more useful, but this also is just incorrect. How does one specify the constraint equations? The algebraic variables?

For a test DAE, use something non-stiff. The ROBER equations for a short time are.

```julia
function f(out,du,u,p,t)
  out[1] = - 0.04u[1]              + 1e4*u[2]*u[3] - du[1]
  out[2] = + 0.04u[1] - 3e7*u[2]^2 - 1e4*u[2]*u[3] - du[2]
  out[3] = u[1] + u[2] + u[3] - 1.0
end

u₀ = [1.0, 0, 0]
du₀ = [-0.04, 0.04, 0.0]
tspan = (0.0,10.0)

using OrdinaryDiffEq
differential_vars = [true,true,false]
prob = DAEProblem(f,du₀,u₀,tspan,differential_vars=differential_vars)
sol = solve(prob,DABDF2())
```

which should be differentiable with Tracker.

src/neural_de.jl
```diff
@@ -167,3 +167,34 @@ function (n::NeuralCDDE)(x,p=n.p)
...
+
+Flux.@functor NeuralDAE
+
+function (n::NeuralDAE)(x,du0=n.du0,p=n.p)
+    function f(u,p,t)
+        vcat([n.differential_vars[i] == 1 ? n.re(p)(u[i]) : n.constraints_model(u[i],p,t) for i in 1:length(n.differential_vars)])
+    end
+    dudt_(u,p,t) = f
+    prob = DAEProblem(dudt_,du0,x,n.tspan,p,differential_vars=n.differential_vars)
+    concrete_solve(prob,n.solver,x,p,n.args...;n.kwargs...)
```
ChrisRackauckas: default to `sensealg = TrackerAdjoint()`

src/neural_de.jl
```diff
@@ -167,3 +167,34 @@ function (n::NeuralCDDE)(x,p=n.p)
...
+            typeof(tspan),typeof(solver),typeof(differential_vars),typeof(args),typeof(kwargs)}(
+            model,constraints_model,p,du0,re,tspan,solver,differential_vars,args,kwargs)
+    end
+end
+
+Flux.@functor NeuralDAE
+
+function (n::NeuralDAE)(x,du0=n.du0,p=n.p)
+    function f(u,p,t)
+        vcat([n.differential_vars[i] == 1 ? n.re(p)(u[i]) : n.constraints_model(u[i],p,t) for i in 1:length(n.differential_vars)])
```
ChrisRackauckas: This isn't quite correct. `n.constraints_model(u,p,t)` should return a vector of constraints. `n.re(p)(u)` returns a vector. Then `n.differential_vars[i] == true` is a better check for decomposing / concatenting.

test/neural_dae.jl
```diff
@@ -0,0 +1,22 @@
+using Flux, DiffEqFlux, OrdinaryDiffEq
```
ChrisRackauckas: this test isn't called until it's added to runtests.jl

src/neural_de.jl
```diff
@@ -167,3 +167,77 @@ function (n::NeuralCDDE)(x,p=n.p)
...
+        new{typeof(p),typeof(model),typeof(constraints_model),typeof(re),
+            typeof(tspan),typeof(solver),typeof(mass_matrix),typeof(args),typeof(kwargs)}(
+            model,constraints_model,p,re,tspan,solver,mass_matrix,args,kwargs)
+    end
+end
+
+Flux.@functor NeuralDAE
+
+function (n::NeuralDAE)(x,p=n.p)
+    function f(du,u,p,t)
```
ChrisRackauckas: use the oop form

src/neural_de.jl
```diff
@@ -167,3 +167,77 @@ function (n::NeuralCDDE)(x,p=n.p)
...
+        nn_out = n.re(p)(u)
+        alg_out = n.constraints_model(u,p,t)
+        du .= vcat(nn_out,alg_out)
+        nothing
+    end
```
ChrisRackauckas:
```suggestion
    function f(u,p,t)
        nn_out = n.re(p)(u)
        alg_out = n.constraints_model(u,p,t)
        vcat(nn_out,alg_out)
    end
```

PR #164, opened 2020-02-18
Trying to train custom RHS's

ranjanan:
I am trying to train an RHS of this form: 
```julia
function rhs2(x,p) 
    half = div(length(p), 2)
    J = dudt1(x, p[1:half]) + dudt2(x, p[half+1:length(p)])
    reshape(J, 3, 3) * x
end
```
It's basically of the form `u' = mat*u` where `mat` depends on the parameters `p`. I had to do this simple patch for the following code to work: 
```julia
n_ode3 = NeuralODE(rhs2, tspan, Rosenbrock23(autodiff = false))
n_ode3(u0, total_p)
```

### DIFF
--- a/src/neural_de.jl
+++ b/src/neural_de.jl
@@ -51,7 +51,7 @@ end
 Flux.@functor NeuralODE
 
 function (n::NeuralODE)(x,p=n.p)
-    dudt_(u,p,t) = n.re(p)(u)
+    dudt_(u,p,t) = n.re(p)(u, p)
     prob = ODEProblem{false}(dudt_,x,n.tspan,p)
     concrete_solve(prob,n.solver,x,p,n.args...;n.kwargs...)
 end


src/neural_de.jl
```diff
@@ -51,7 +51,7 @@ end
 Flux.@functor NeuralODE
 
 function (n::NeuralODE)(x,p=n.p)
-    dudt_(u,p,t) = n.re(p)(u)
+    dudt_(u,p,t) = n.re(p)(u, p)
```
ChrisRackauckas: This isn't correct since Flux models are one-argument functions.

ChrisRackauckas:
You probably shouldn't be using NeuralODE for this. This is a pretty custom setup and you'll want to create a struct for it.

src/neural_de.jl
ranjanan: But 
```
m = FastChain(FastDense(3, 9, tanh))
```
`m` here is a two argument function right? And that's the input being passed to `NeuralODE`. So I made `rhs2` a two argument function as well.

ranjanan: Note that `n.re(p)` returns `rhs2`

ChrisRackauckas:
FastChain is a different dispatch: https://github.com/ranjanan/DiffEqFlux.jl/blob/0a6830aacbfb0433f9e565df3b587388149a0478/src/neural_de.jl#L59

ranjanan:
Ah right. Then `n_ode3 = NeuralODE(FastChain(rhs2), tspan, Rosenbrock23(autodiff = false))` should work right?

ChrisRackauckas:
Yes

PR #177, opened 2020-03-04
Setup NLopt optimizers with sciml_train

### LINKED ISSUE #161: Setup NLopt optimizers with sciml_train
It would be interesting to have NLopt work on the same interface. It would give access to a bunch of different schemes, many global, and check how derivative-free methods do.

abhigupta768:
Fixes #161


src/DiffEqFlux.jl
```diff
@@ -2,7 +2,7 @@ module DiffEqFlux
 
 using DiffEqBase, Tracker, DiffResults, DiffEqSensitivity, ForwardDiff,
       Flux, Requires, Adapt, LinearAlgebra, RecursiveArrayTools, Juno, Optim,
-      StaticArrays, Base.Iterators
+      StaticArrays, Base.Iterators, NLopt
```
ChrisRackauckas: do an import here

src/train.jl
```diff
@@ -202,11 +202,33 @@ function sciml_train(loss, θ, opt::Optim.AbstractOptimizer, data = DEFAULT_DATA
...
+  function nlopt_grad!(θ,grad)
+    _x,lambda = Flux.Zygote.pullback(θ) do θ
+      x = loss(θ)
+      first(x)
+    end
+
+    if length(grad) > 0 
+      grad .= first(lambda(1))
+    end
+    println(_x)
```
ChrisRackauckas: ?

src/train.jl
```diff
@@ -202,11 +202,33 @@ function sciml_train(loss, θ, opt::Optim.AbstractOptimizer, data = DEFAULT_DATA
...
+    return _x
+  end
+
+  min_objective!(opt, nlopt_grad!)
+
+  NLopt.optimize(opt, θ)
+end
```
ChrisRackauckas: the callback isn't used anywhere

test/layers_sciml.jl
```diff
@@ -104,3 +104,9 @@ loss2 = loss_adjoint(pmin.minimizer)
 pmin = DiffEqFlux.sciml_train(loss_adjoint, p, Fminbox(BFGS(initial_stepnorm = 0.01)), lower_bounds = [0.0 for i in 1:4], upper_bounds = [5.0 for i in 1:4], cb = cb)
 loss2 = loss_adjoint(pmin.minimizer)
 @test 10loss2 < loss1
+
+opt = Opt(:LD_MMA, 4)
+maxiters!(opt, 100)
```
ChrisRackauckas: we should at least commonize this part of the interface

ChrisRackauckas:
Nice. Looks good. Now can we set this up via Requires.jl so that we don't have to take on the binary dependency (it can be hardware-limiting). See https://github.com/JuliaDiffEq/DiffEqFlux.jl/blob/master/src/DiffEqFlux.jl#L10-L19 . Move init to a new file and do a require that adds these dispatches.

src/DiffEqFlux.jl
```diff
@@ -65,6 +55,9 @@ Flux.Zygote.grad_mut(d::IdDict) = IdDict()
 include("train.jl")
 include("fast_layers.jl")
 include("neural_de.jl")
+include("require.jl")
+
+__init__()
```
ChrisRackauckas: don't need to run it.

ChrisRackauckas:
```suggestion
```

PR #180, opened 2020-03-08
[WIP] Add progress bar to sciml_train

abhigupta768:


Vaibhavdixit02:
LGTM.

src/train.jl
```diff
@@ -73,9 +73,15 @@ function sciml_train(loss, _θ, opt, _data = DEFAULT_DATA;
     data = _data
   end
 
+  progress && @logmsg(LogLevel(-1),
+  "Training",
+  _id = :DiffEqFlux,
+  message="init params = "*string(ps),
```
ChrisRackauckas: no need to print this. Instead have the message be the current loss?

Vaibhavdixit02: Hmm, I didn't see the message at all.

ChrisRackauckas: if you scroll over the progress bar it should show.

PR #181, opened 2020-03-09
support second order

ChrisRackauckas:


ChrisRackauckas:
Needs https://github.com/JuliaNLSolvers/Optim.jl/pull/793 to work well with Float32s, so I hope @pkofod will get around to it, but otherwise it's good to go.

PR #184, opened 2020-03-12
Provide default progress loggers and use new ProgressLogging API

devmotion:
This PR fixes https://github.com/JuliaDiffEq/DiffEqFlux.jl/issues/158 and adds support for the new (unreleased) support of default progress loggers in DiffEqBase to DiffEqFlux. In my local tests, progress bars were displayed in the terminal successfully (but maybe you want to disable the progress bars in the CI tests since it seems to fill up the logs IIRC).

As mentioned by @tkf in https://github.com/JuliaDiffEq/DiffEqFlux.jl/issues/158#issuecomment-586768555, the problem with the current approach of setting up the progress logs (and also in https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/183) is that the progress bar can't distinguish between a crashed and a slow progress. To get this right, the PR makes use of `ProgressLogging.@withprogress` and the new ProgressLogging API. In this way also a UUID is used as id automatically. However, since `ProgressLogging.@withprogress` automatically initializes the progress bar before the first manual progress update (in https://github.com/JunoLab/ProgressLogging.jl/blob/b64333df0f4d2e780387af01867cccb7460da33d/src/ProgressLogging.jl#L326), a macro helps to use it only if progress output is desired. Otherwise, a progress bar might be initialized and finalized after the computations even though the user specified `progress = false`.


devmotion:
It seems Travis tests fail due to the same error as on master.

devmotion:
Should the progress bars be disabled in the tests?

ChrisRackauckas:
The error on master: @pkofod did you forget to tag Optim?

PR #187, opened 2020-03-14
auto-dualize the NeuralDE input

ChrisRackauckas:
Otherwise users would need to do things like `nODE(eltype(θ).(x), θ)` which could get confusing, so we might as well just directly handle it since we know it will come up.


ChrisRackauckas:
Supersceded by changes to DiffEqBase.

PR #191, opened 2020-03-17
fix and better test neural SDE on FastChain

ChrisRackauckas:
Was pulling the same parameter set twice instead of grabbing the two halves


ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/DiffEqFlux.jl/issues/189

Issue #170, opened 2020-02-20
Store and return the current best loss from sciml_train on Flux.jl optimizers


ChrisRackauckas:
By default, some Flux optimizers don't always move forward due to stochastic optimization. In these cases, you can have the loss jiggle, so you don't necessarily return the parameters with the lowest loss if you return the last parameters. However, we can store the current best and then manually return the true best at the end of the optimization.

dhairyagandhi96:
Seems like a good utility to have in Flux, @MikeInnes?

MikeInnes:
This just seems like a way to overfit to the test set. The fluctuations after an optimiser has converged likely just represent noise caused by the minibatches, and don't reflect true performance (i.e. on new data rather than just the test set).

Besides that there's the technical issue of having to make copies of a (potentially large) model. So I don't think we could or should do this by default, but perhaps if someone can design a separate utility and show that it's sound we could consider it.

PR #195, opened 2020-03-22
Save best solution rather than last one in sciml_train

ChrisDeGrendele:
This PR saves the best solution from sciml_train rather than the last one produced. This will be helpful for things like stochastic loss functions. This is only first of three sciml_train, I was hoping to see if I am on the right track.
Fixes Issue #170

### DIFF
--- a/src/train.jl
+++ b/src/train.jl
@@ -64,10 +64,13 @@ The keyword arguments are as follows:
 - `maxiters`: Specifies the maximum number of iterations for the optimization.
   Required if a Flux optimizer is chosen and no data iterator is given, otherwise
   defaults to infinite.
+- `save_best`: Specifies whether you want the best solution, corresponding to the
+  lowest loss function value. If false, the last solution found will be returned.
 """
 function sciml_train(loss, _θ, opt, _data = DEFAULT_DATA;
                      cb = (args...) -> false,
-                     maxiters = get_maxiters(data), progress=true)
+                     maxiters = get_maxiters(data),
+                     progress=true, save_best=true)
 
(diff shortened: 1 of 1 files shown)


src/train.jl
```diff
@@ -86,7 +86,9 @@ function sciml_train(loss, _θ, opt, _data = DEFAULT_DATA;
 
   t0 = time()
 
-  local x
+  local x, min_err
+  min_err = 999999 #dummy variables
```
ChrisRackauckas: typemax(eltype(_θ))

src/train.jl
```diff
@@ -102,6 +104,15 @@ function sciml_train(loss, _θ, opt, _data = DEFAULT_DATA;
       msg = @sprintf("loss: %.3g", x[1])
       progress && ProgressLogging.@logprogress msg i/maxiters
       update!(opt, ps, gs)
+
+      if x < min_err  #found a better solution
```
ChrisRackauckas: we should have it be an option to do this.

titus-pinta:
Sorry if this is a mistake, or if I do not post in the correct place, but there seems to be a mistake. If I return more than 1 variable from my loss function in order to use the callback, the function first(x) on line 130 of train.jl will try to compare a pair of objects and will throw an error. I think that replacing it with first(x[1]) should solve the problem.
Sorry if I miss something obvious here!

ChrisDeGrendele:
@titus-pinta Hey, good catch. I was just able to reproduce your error, so I can have a fix pushed in a bit. Thanks!

PR #196, opened 2020-03-24
Add BlackBoxOptim dispatch to sciml_train

abhigupta768:


src/train.jl
```diff
@@ -308,3 +308,26 @@ function sciml_train(loss, θ, opt::Optim.AbstractConstrainedOptimizer,
...
                          f_calls_limit = maxiters))
 end
+
+
+function sciml_train(loss, opt::Symbol = :adaptive_de_rand_1_bin, data = DEFAULT_DATA;lower_bounds, upper_bounds,
+                      cb = (args...) -> (false), maxiters = get_maxiters(data), kwargs...)
+  local x, cur, state
+  cur,state = iterate(data)
+
+  function _cb(trace)
```
Vaibhavdixit02: The callback is not being passed here?

Vaibhavdixit02:
I think you should add Evolutionary.jl dispatch to the same PR. Let's just close the issue in a go 🙂

abhigupta768:
I took a look at Evolutionary.jl, they don't have an uniform way of calling optimize with different algorithms. So we would need dispatch to each method independently, not sure if that is desirable?

Vaibhavdixit02:
@ChrisRackauckas what do you think?

ChrisRackauckas:
Dispatching on symbol isn't too safe... I would prefer it to be like `BBO(:method)` or something like that.

src/train.jl
```diff
@@ -308,3 +308,17 @@ function sciml_train(loss, θ, opt::Optim.AbstractConstrainedOptimizer,
...
+                      maxiters = get_maxiters(data), kwargs...)
+  local x, cur, state
+  cur,state = iterate(data)
+
+  _loss = function (θ)
+    x = loss(θ,cur...)
+    first(x)
+  end
+
+  BlackBoxOptim.bboptimize(_loss;Method = opt, SearchRange = [(lower_bounds[i], upper_bounds[i]) for i in 1:length(lower_bounds)], kwargs...)
```
ChrisDeGrendele: @ChrisRackauckas This should then be converted to an Optim.jl optimum solution? To keep `sciml_train` consistent

ChrisRackauckas: yes

src/train.jl
```diff
@@ -308,3 +308,48 @@ function sciml_train(loss, θ, opt::Optim.AbstractConstrainedOptimizer,
            Optim.Options(extended_trace=true,callback = _cb,
                          f_calls_limit = maxiters))
 end
+
+struct BBO
+  method::Symbol = :adaptive_de_rand_1_bin
```
ChrisRackauckas: I don't think you can set defaults like that.

src/train.jl
```diff
@@ -309,8 +309,11 @@ function sciml_train(loss, θ, opt::Optim.AbstractConstrainedOptimizer,
                          f_calls_limit = maxiters))
 end
 
+struct BBO
+  method::Symbol = :adaptive_de_rand_1_bin         
+end
```
ChrisRackauckas:
```suggestion
  method::Symbol         
end
BBO() = BBO(:adaptive_de_rand_1_bin)

```

Vaibhavdixit02:
@ChrisRackauckas I can't see the Travis run for this anymore, Travis needs some update?

Issue #199, opened 2020-03-26
BBO breaks


ranjanan:
```julia
using DiffEqFlux, OrdinaryDiffEq, Flux, Optim, Plots                                                                                                                                        
                                                                                                                                                                                             
 u0 = Float32[2.; 0.]                                                                                                                                                                        
 datasize = 30                                                                                                                                                                               
 tspan = (0.0f0,1.5f0)                                                                                                                                                                       
                                                                                                                                                                                              function trueODEfunc(du,u,p,t)                                                                                                                                                                           true_A = [-0.1 2.0; -2.0 -0.1]                                                                                                                                                          
     du .= ((u.^3)'true_A)'                                                                                                                                                                   end
                                                                                                                                                                                         
 t = range(tspan[1],tspan[2],length=datasize)                                                                                                                                                
 prob = ODEProblem(trueODEfunc,u0,tspan)                                                                                                                                                     
 ode_data = Array(solve(prob,Tsit5(),saveat=t))                                                                                                                                              
                                                                                                                                                                                             
 dudt2 = FastChain((x,p) -> x.^3,                                                                                                                                                            
             FastDense(2,50,tanh),                                                                                                                                                           
             FastDense(50,2))                                                                                                                                                                
 n_ode = NeuralODE(dudt2,tspan,Tsit5(),saveat=t)                                                                                                                                             
                                                                                                                                                                                             
 function predict_n_ode(p)                                                                                                                                                                   
   n_ode(eltype(p).(u0),p)                                                                                                                                                                   
 end                                                                                                                                                                                         
                                                                                                                                                                                             
 function loss_n_ode(p)                                                                                                                                                                      
     pred = predict_n_ode(p)                                                                                                                                                                 
     loss = sum(abs2,ode_data .- pred)                                                                                                                                                       
     loss,pred                                                                                                                                                                               
 end                                                                                                                                                                                         
                                                                                                                                                                                             
 loss_n_ode(n_ode.p) # n_ode.p stores the initial parameters of the neural ODE                                                                                                               
                                                                                                                                                                                             
 cb = function (p,l,pred;doplot=false) #callback function to observe training                                                                                                                
   display(l)                                                                                                                                                                                
   # plot current prediction against data                                                                                                                                                    
   if doplot                                                                                                                                                                                 
     pl = scatter(t,ode_data[1,:],label="data")                                                                                                                                              
     scatter!(pl,t,pred[1,:],label="prediction")                                                                                                                                             
[...]

PR #200, opened 2020-03-26
Change BBO interface

ranjanan:
I think the API should be more like this. The standard error will be thrown when users don't supply lower and upper bounds. Additonally, let us take in the parameters and do nothing with them.

### DIFF
--- a/src/train.jl
+++ b/src/train.jl
@@ -315,7 +315,7 @@ end
 
 BBO() = BBO(:adaptive_de_rand_1_bin)
 
-function sciml_train(loss, opt::BBO = BBO(), data = DEFAULT_DATA;lower_bounds, upper_bounds,
+function sciml_train(loss, _θ, opt::BBO, lower_bounds, upper_bounds, data = DEFAULT_DATA;
                       maxiters = get_maxiters(data), kwargs...)
   local x, cur, state
   cur,state = iterate(data)


ChrisRackauckas:
we need to make sure that all dispatches handle box constraints the same.

ranjanan:
I think they can all use this interface.

ChrisRackauckas:
Okay, then they should all change in the same PR.

Vaibhavdixit02:
I think having the bounds ask kwargs is a better interface, it doesn't leave room for any ambiguity wrt error in passing them which might come up in case of positional args?

ranjanan:
But @Vaibhavdixit02 they're compulsory arguments that must be passed so they shouldn't be kwargs.

ChrisRackauckas:
Mandatory kwargs are a thing.

ranjanan:
Sure, but if you do have mandatory kwargs, you'd have to have specific error handling in place if they aren't passed right?

ChrisRackauckas:
You just don't give it a default value. 

https://docs.julialang.org/en/v1/manual/functions/#Keyword-Arguments-1

>If a keyword argument is not assigned a default value in the method definition, then it is required: an UndefKeywordError exception will be thrown if the caller does not assign it a value:

ChrisRackauckas:
If you're not hitting that, then what's going on is that it's somehow hitting the wrong dispatch (probably because Flux has issues)

ranjanan:
I'm ok with keeping these mandatory kwargs, but can we keep the `\theta` second argument and do nothing with it for consistency? Otherwise, there's a chance I get the original error in #199

ChrisRackauckas:
I think the right thing to do is to tell Flux that no, it is not "better" to have no abstract type on your optimizers, and make it so Flux has to obey dispatch so that it matches correctly.

abhigupta768:
>but can we keep the \theta second argument and do nothing with it for consistency?

I think this is a good idea, I had meant to do the same but it slipped out of my attention

PR #204, opened 2020-03-29
Change of Dimensions for Neural ODE issue #111

### LINKED ISSUE #111: Suggestion:  Change dimension order on NeuralODE to match convention of ML batch
Issue:
NeuralODE when driven by a batch input will output an array in which the last index is the time index not the batch index.  This is different than the standard convention in all other ML libraries including Flux, TensorFlow, Pytorch, and Chainer.
 In Flux.conv the batch output dimensions are
( Data-dimension, channel index, batch index)

the time series output from a NeuralODE is analagous to the Channel dimension.  Thus it would be very logical for the output of a nerualODE to place the time dimension second to the last, so that the batch index can be the last dimension.

Usecase
In addition to simply being logical and following the ML convention there are other benefits.

When generting a batch of labels for training using a typical loss function one will generate this batch, often, from the output of some other neuralNet fed the batchinput.  That will put the batch index on the end.  Alternatively one might generte the batch Labels by hcat-ing a set of individual results together.
in either case the batch index is the last dimension.

If the neuralODE had the same dimenstional order then a loss function written for the single label case will also work for the batch label case because the last dimension will broadcast when subtrating the neuralODE batch output from the batch labels.

While you could do a permute on the last two indicies of the neuralODE batch output this is icky for two reasons.  First it means the loss function has to detect if it's operating on a batch or a single case and apply the permute in the batch case.  Second for a lazy permute, the memory order is not contiguous and so in the event of large tensor sizes inefficient in the memory page retrieval orders or caching.

Suggested Path:
add a key word to NeuralODE called "UseMLBatchOrder= true" to set the preferred output.  Intially hack this in just with a dimenstion permute.  Later on if this turns out to be the preference everyone likes you make the memory storage order concrete in this order and remove the need for the permute.

titus-pinta:
This tries to change the order of dimensions from NeuralODE to have as the last index the batch id. For instance if the Neural ODE solver would normaly return a (5, 5) array at each t, than with this option it will return a (5, 5, 1, 1) array, where the second to last item is time and the last item is batch id.
Here is some code example https://github.com/titus-pinta/DiffEqFlux.jl/blob/master/src/minibatch_conv_test.jl
I have decided to implement it in this way in order to not break any existing code that usees NeuralODE. Also I wanted to avoid adding an extra if in the original code for (n::NeuralODE)(x,p) in order to, similarly, not affect the code that uses DiffEqFlux.
I am sorry if this is not well formated, or is not in the right place, this is my first commit to something related to Julia. Also I am new to Julia, so the code may not be very well written. Also sorry if I did not understood the problem well, and this is not actually the solution to the correct question.

Fixes #111


titus-pinta:
This is for issue #111, but I don't think that I can link an issue to this pull request.

ChrisRackauckas:
>This is for issue #111, but I don't think that I can link an issue to this pull request.

You have to put it in the original message for it to link to the issue and auto-close, so I edited it.

src/neural_de.jl
```diff
@@ -48,6 +48,36 @@ struct NeuralODE{M,P,RE,T,S,A,K} <: NeuralDELayer
...
+        new{typeof(model),typeof(p),typeof(re),
+            typeof(tspan),typeof(solver),typeof(args),typeof(kwargs)}(
+            n,model,p,re,tspan,solver,args,kwargs)
+    end
+end
+
+Flux.@functor NeuralODEBatchConv
+function (n::NeuralODEBatchConv)(x, args...)
+    pred = n.n(x, args...)
+    reshape(Array(pred), (size(pred)..., 1))
```
ChrisRackauckas: I don't think this is right: you'll need to permute the dimensions to move it to the last one.

ChrisRackauckas:
This is a few versions behind by now, so closing.

PR #209, opened 2020-04-02
Default allow_f_increases=true

YingboMa:


ChrisRackauckas:
@ranjanan opinion here?

ranjanan:
Fantastic!

ChrisRackauckas:
From the tests, it seems that if we do this the optimization never haults, so we may just want to make this a well-known argument but not a default? Or if it is a default, require maxiters?

ranjanan:
Let's not make it a default. Yes, if it's supplied, you must also enter a maxiters. Using this option can be a recommended experiment. Worse case the loss blows up and we can discard it.

ChrisRackauckas:
Losses going to 1e-82 is absolutely fantastic though :P

PR #215, opened 2020-04-06
Make lower_bounds/upper_bounds mandatory kwargs

ChrisDeGrendele:
- Moving lower_bounds and upper_bounds to mandatory kwargs when needed in 2/4 sciml's. 
- Pass _theta to the blackbox sciml even though it won't use it. 

This PR allows us to call all 4 sciml_train functions by only swapping opt() making them more uniform.

### DIFF
--- a/src/train.jl
+++ b/src/train.jl
@@ -272,15 +272,10 @@ end
 
 
 function sciml_train(loss, θ, opt::Optim.AbstractConstrainedOptimizer,
-                     data = DEFAULT_DATA;
+                     data = DEFAULT_DATA; lower_bounds, upper_bounds,
                      cb = (args...) -> (false), maxiters = get_maxiters(data), kwargs...)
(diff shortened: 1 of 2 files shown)


src/train.jl
```diff
@@ -273,11 +273,15 @@ end
 
 function sciml_train(loss, θ, opt::Optim.AbstractConstrainedOptimizer,
                      data = DEFAULT_DATA;
-                     lower_bounds, upper_bounds,
-                     cb = (args...) -> (false), maxiters = get_maxiters(data))
+                     cb = (args...) -> (false), maxiters = get_maxiters(data), kwargs...)
   local x, cur, state
   cur,state = iterate(data)
 
+  KD= Dict( kwargs ) #Mandatory kwargs
```
ChrisRackauckas, line 280: just use the standard mandatory kwargs handling.

PR #201, opened 2020-03-26
Add callback kwarg to BlackBoxOptim dispatch

abhigupta768:

PR #216, opened 2020-04-06
Changed Sciml Interface

ChrisDeGrendele:
Sciml_trains now have a common interface allowing for uniform code. All can be called with the same arguments just by swapping opt.


Vaibhavdixit02:
This is already handled in #201?

ChrisDeGrendele:
Ah ya, it mostly is. I didn't see that.

PR #217, opened 2020-04-07
Add TikTak global optimization to sciml_train

ranjanan:


src/train.jl
```diff
@@ -355,3 +355,51 @@ function sciml_train(loss, opt::BBO = BBO(), data = DEFAULT_DATA;lower_bounds, u
                                         NaN,
                                         bboptre.elapsed_time)
 end
+
+function sciml_train(loss, opt::TikTak, data = DEFAULT_DATA;lower_bounds, upper_bounds, local_method,
```
ChrisRackauckas, line 359: initial condition for uniformity?

src/train.jl
```diff
@@ -355,3 +355,51 @@ function sciml_train(loss, opt::BBO = BBO(), data = DEFAULT_DATA;lower_bounds, u
                                         NaN,
                                         bboptre.elapsed_time)
 end
+
+function sciml_train(loss, opt::TikTak, data = DEFAULT_DATA;lower_bounds, upper_bounds, local_method,
+                      maxiters = get_maxiters(data), kwargs...)
```
ChrisRackauckas, line 360: any way to do callback support?

src/train.jl
ranjanan: Sure

src/train.jl
ranjanan: No, there does not seem to be any callback support.

ChrisRackauckas:
test?

ChrisRackauckas:
test it in the neural network context like the others: https://github.com/SciML/DiffEqFlux.jl/blob/master/test/layers_sciml.jl#L91-L144 . It would probably be good to refactor that whole deal to its own file though.

ChrisRackauckas:
seems like overkill for 3.

PR #220, opened 2020-04-08
Add QuadDIRECT global optimizer 

ranjanan:


src/require.jl
```diff
@@ -168,5 +168,54 @@ function __init__()
                                                 t1 - t0)
         end
     end
+    @require QuadDIRECT = "dae52e8d-d666-5120-a592-9e15c33b8d7a"
+        export QuadDIRECT
```
ChrisRackauckas, line 172: why export?

ranjanan: The Optim solvers are exported. So I thought I'd export this as well. We don't need to though. Would you prefer to not export this?

ranjanan:
@ChrisRackauckas QuadDIRECT is registered in a different registry (HolyLabs). How do I make it available in the tests? Do I add a line in Travis about adding HolyLabs as a registry?

src/require.jl
ChrisRackauckas: if you have to use requires, then `using QuadDIRECT` is already done on the user side, so it's all already there.

src/require.jl
```diff
@@ -168,7 +168,7 @@ function __init__()
                                                 t1 - t0)
         end
     end
-    @require QuadDIRECT = "dae52e8d-d666-5120-a592-9e15c33b8d7a"
+    @require QuadDIRECT = "dae52e8d-d666-5120-a592-9e15c33b8d7a" begin
         export QuadDIRECT
         struct QuadDIRECT
```
ChrisRackauckas, line 173: is that not going to cause issues being the same name as the package?

.travis.yml
```diff
@@ -10,6 +10,9 @@ julia:
 #    - julia: nightly
 notifications:
   email: false
+script: 
+ - julia -e 'using Pkg; pkg"registry add git@github.com:HolyLab/HolyLabRegistry.git"'
```
ChrisRackauckas, line 14: ugh that's ugly.

PR #223, opened 2020-05-01
analytical tgrad for neural ODEs

ChrisRackauckas:


ChrisRackauckas:
@ranjanan it looks like Travis hasn't worked since your last PR

PR #227, opened 2020-05-04
Readme update

Emmanuel-R8:
This is a bulk update of the README. The changes are:

- addition of Gitpod
- Code formatting. While we often have different tastes in formatting, one thing we all agree on is that it has to be consistent.  Some functions have been cosmetically rewritten to be much more explicit about what is happening.
- Added Latex for the Lotka-Volterra equations. But it seems that Github doesn't render them without browser extensions.
- Naming: Minimise resuse of the same names. Aside from the names of the callback functions and some variables, all names are now specific to each example.
- Debugging: all functions run fine.
- Visuals: create gif's from plots accumulated in callbacks and insert them in the document. These gifs are stored in a new `assets` folder (DifferentialEquations.jl also has an `assets` folder, it so it seemed OK to do the same here)

Aside from those, the structure of the document and the examples are unchanged.


README.md
```diff
@@ -208,102 +273,155 @@ supported. For example, we can build a layer with a delay differential equation
...
 end
-loss_dde(p) = sum(abs2,x-1 for x in predict_dde(p))
+
+loss_dde(p) = sum(abs2, x-1 for x in predict_dde(p))
 loss_dde(p)
 ```
 
-Notice that we chose `sensealg=TrackerAdjoint()` to utilize the Tracker.jl
+```
+369.9247935840374
```
Vaibhavdixit02, line 308: I am not sure adding the output values manually is a good idea

Vaibhavdixit02:
I couldn't figure out what the Gitpod addition is doing, can you explain the motivation for that in more detail?

README.md
Emmanuel-R8: I left them in to give the reader a sense of what to expect when executing. 

As a dynamic content document format, Github markdown is not good enough. I thought about writing something in Julia markdown to generate this file. Maybe that would be a better solution.

Emmanuel-R8:
Gitpod gived me real time rendering of the markdown and reproducible environment. But if consensus is to remove, no problem.

Vaibhavdixit02:
> But if consensus is to remove, no problem.
I wasn't suggesting that. Just trying to understand the implications better.

Can you resolve the conflict so that Travis runs?

README.md
```diff
@@ -1,38 +1,51 @@
+[![Gitpod Ready-to-Code](https://img.shields.io/badge/Gitpod-Ready--to--Code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/SciML/DiffEqFlux.jl)
+
 # DiffEqFlux.jl
 
 [![Join the chat at https://gitter.im/JuliaDiffEq/Lobby](https://badges.gitter.im/JuliaDiffEq/Lobby.svg)](https://gitter.im/JuliaDiffEq/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![Gitpod Ready-to-Code](https://img.shields.io/badge/Gitpod-Ready--to--Code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/SciML/DiffEqFlux.jl)
 [![Build Status](https://travis-ci.org/SciML/DiffEqFlux.jl.svg?branch=master)](https://travis-ci.org/SciML/DiffEqFlux.jl)
 [![Build status](https://ci.appveyor.com/api/projects/status/e5a9pad58ojo26ir?svg=true)](https://ci.appveyor.com/project/ChrisRackauckas/diffeqflux-jl)
-[![GitlabCI](https://gitlab.com/sciml/DiffEqFlux-jl/badges/master/pipeline.svg)](https://gitlab.com/sciml/DiffEqFlux-jl/pipelines)
+[![GitlabCI](https://gitlab.com/juliadiffeq/DiffEqFlux-jl/badges/master/pipeline.svg)](https://gitlab.com/juliadiffeq/DiffEqFlux-jl/pipelines)
```
ChrisRackauckas, line 9: looks like this needs to be rebased to master

README.md
```diff
@@ -1,38 +1,51 @@
...
 
-DiffEqFlux.jl fuses the world of differential equations with machine learning
-by helping users put diffeq solvers into neural networks. This package utilizes
+DiffEqFlux.jl fuses the world of differential equations with machine learning by
+helping users put diffeq solvers into neural networks. This package utilizes
```
ChrisRackauckas, line 12: formatting changes should be kept in a separate PR

README.md
```diff
@@ -94,76 +107,123 @@ For an overview of what this package is for,
...
+sol_ode = solve(prob_ode, Tsit5())
+
+# Plot the solution
 using Plots
-plot(sol)
+savefig(plot(sol_ode), "assets/LV_ode.png")
 ```
 
-![LV Solution Plot](https://user-images.githubusercontent.com/1814174/51388169-9a07f300-1af6-11e9-8c6c-83c41e81d11c.png)
+![LV Solution Plot](assets/LV_ode.png)
```
ChrisRackauckas, line 149: Assets should not be added to the repo for this. That will cause repo bloat.

README.md
```diff
@@ -208,102 +273,155 @@ supported. For example, we can build a layer with a delay differential equation
...
+its derivative function. Thus for example, with the multilayer perceptron neural
+network `Chain(Dense(2, 50, tanh), Dense(50, 2))`, the best way to define a
+neural ODE by hand would be to use non-mutating adjoints, which looks like:
 
-```julia
-p,re = Flux.destructure(model)
-dudt_(u,p,t) = re(p)(u)
-prob = ODEProblem(dudt_,x,tspan,p)
-my_neural_ode_prob = concrete_solve(prob,Tsit5(),u0,p,args...;kwargs...)
+```
```
ChrisRackauckas, line 401: should be marked as Julia code for syntax highlighting

ChrisRackauckas:
There's a lot of good here, but there are a lot of changes to be fixed.

>Gitpod gived me real time rendering of the markdown and reproducible environment. But if consensus is to remove, no problem.

I'd rather keep this simple. In fact, the README is soon going to be replaced with real documentation.

>DifferentialEquations.jl also has an assets folder, it so it seemed OK to do the same here

We stopped doing that because it bloats the repo. The repo size of DifferentialEquations.jl is 100mb because of previously doing this, which is why we no longer accept that and moved the docs to DiffEqDocs.jl. Assets like that stay in history, so you can never delete them without changing the SHAs which means it would break registered release tags. Moral of the story is, we shouldn't add assets like that, and instead drop them into the repo's issues/PRs and then link to them. The PR should completely remove them from its history, i.e. adding a commit that deletes them is not enough to remove it from memory and instead it needs to be squashed over.

> Added Latex for the Lotka-Volterra equations. But it seems that Github doesn't render them without browser extensions.

That'll be fixed by putting it in real Documenter.jl docs.

> Naming: Minimise resuse of the same names. Aside from the names of the callback functions and some variables, all names are now specific to each example.

Nice, though will be less necessary when they are all different pages.

nisarhassan12:
> I couldn't figure out what the Gitpod addition is doing, can you explain the motivation for that in more detail?

Thanks @Emmanuel-R8 for the Pr.

Gitpod is an online IDE for GitHub and GitLab that automates dev environments for any project with a single click.

In the case of this project with a single click Gitpod will launch a workspace and automatically:

- clone the `DiffEqFlux.jl` repo.
- install Julia and it's dependencies.
- and install VS Code extensions related to Julia support.

Gitpod is free for OSS and can be used to simplify code contributions + reviews. A lot of popular OSS projects out there are already using it simplify contributors onboarding some of them are:

- https://github.com/freeCodeCamp/freeCodeCamp
- https://github.com/facebook/docusaurus
- https://github.com/mozilla/pdf.js/

You can find a brief list of them at https://contribute.dev

image

This is how it looks:

image

You can give it a try on @Emmanuel-R8 fork of this repo via the following link:

https://gitpod.io/#https://github.com/Emmanuel-R8/DiffEqFlux.jl/tree/README-update

.gitpod.yml
```diff
@@ -0,0 +1,9 @@
+image:
+  file: .gitpod.Dockerfile
+tasks:
+- command: julia --version
+vscode:
+  extensions:
+    - julialang.language-julia
+    - colinfang.markdown-julia
+    - singularitti.vscode-julia-formatter
```
nisarhassan12, line 9: These extensions are not getting installed you have to give the following ids in order for them to be installed.

```suggestion
  extensions:
    - singularitti.vscode-julia-formatter@0.1.0:RmhU34gIXDaV/d9sKVGQWg==
    - colinfang.markdown-julia@0.0.1:7o0taaLFuuX28a/ksp1m7g==
    - julialang.language-julia@0.15.29:6iA33XMcDFiLuabO41qB5A==
```

Emmanuel-R8:
Given Chris's comment, it sounds like using this PR as a starting point for Documenter.jl docs is a better way to spend time than finalising it.

Which sub-repo has a existing docs that I could get inspiration from?

ChrisRackauckas:
ModelingToolkit.jl might be a good model, or DataDrivenDiffEq

PR #228, opened 2020-05-06
Documenter

Emmanuel-R8:
This is a translation of the previous pull request to use documenter.

One issue is the neural  network SDE example. The last optimisation is far too slow and I didn't allow it to run (about 4/5 hours on my laptop)


README.md
```diff
@@ -1,10 +1,5 @@
 # DiffEqFlux.jl
 
-[![Join the chat at https://gitter.im/JuliaDiffEq/Lobby](https://badges.gitter.im/JuliaDiffEq/Lobby.svg)](https://gitter.im/JuliaDiffEq/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
```
ChrisRackauckas, line 3: CI badges and such should not be removed.

.gitpod.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM gitpod/workspace-full
```
ChrisRackauckas, line 1: I'm not sure what gitpod adds here.

ChrisRackauckas:
>One issue is the neural network SDE example. The last optimisation is far too slow and I didn't allow it to run (about 4/5 hours on my laptop)

Yes, it takes awhile. I don't expect that one to be ran as part of the doc building. In fact, I usually don't expect most examples to be built in the docs. However, a lot of the DiffEqFlux ones are fast enough that it's probably okay.

docs/src/API.md
```diff
@@ -0,0 +1,36 @@
+# API Documentation
+
+## Neural DE Layer Functions
+
+- `NeuralODE(model, tspan, solver, args...; kwargs...)` defines a neural ODE
```
ChrisRackauckas, line 5: it may be good to do these as docstrings.

.gitpod.Dockerfile
Emmanuel-R8: I'll  remove.

README.md
Emmanuel-R8: I'll revert.

Emmanuel-R8: But I changed the gitlab repo to juliadiffeq. SciML doesn't seem to exist.

docs/src/API.md
Emmanuel-R8: I'll keep that for another commit.

.gitlab-ci.yml
```diff
@@ -1,59 +0,0 @@
-image: "julia:1"
```
ChrisRackauckas, line 1: please don't remove our GPU CI. We know it's down right now (since we can't get on campus), but that isn't a reason to delete it.

.travis.yml
```diff
@@ -10,6 +10,15 @@ julia:
 #    - julia: nightly
 notifications:
   email: false
+<<<<<<< HEAD
+<<<<<<< HEAD
+=======
+=======
+>>>>>>> ee0fb39d33f9a4065a18b58fee0f12783d8a74e2
```
ChrisRackauckas, line 17: merge error

README.md
ChrisRackauckas: oh yes, the Gitter is still JuliaDiffEq

Emmanuel-R8:
Reverted gitlab-CI. Cleaned-up the mess in Travis.

ChrisRackauckas:
Thanks, it's looking great! I don't think we're quite ready to release the docs because I want to add more reference documentation, but this is at least a great foundation to finish the docs from. Thanks a lot.

@kanav99 we should look into getting deployment ready here.

kanav99:
Yup, should I do it today or wait for final release docs?

ChrisRackauckas:
whenever you have time. We'll need to get a subdomain setup for it to show anyways, so there's no rush.

PR #234, opened 2020-05-12
Add MLSL global optimization

AStupidBear:


src/require.jl
```diff
@@ -85,6 +85,15 @@ function __init__()
           NLopt.upper_bounds!(opt, upper_bounds)
           NLopt.maxeval!(opt, maxeval)
 
+          if nstart > 1
+            localopt = opt
+            opt = NLopt.Opt(:G_MLSL_LDS, length(lower_bounds))
```
Vaibhavdixit02, line 90: I would expect this to be passed in as the `opt`, it would be better to have a `localopt` kwarg and use that. This also hard codes the name, in the NLopt docs I can see Augmented Lagrangian algorithm is another algorithm that would work this way.

src/require.jl
```diff
@@ -85,6 +85,15 @@ function __init__()
...
+            NLopt.lower_bounds!(opt, lower_bounds)
+            NLopt.upper_bounds!(opt, upper_bounds)
+            NLopt.local_optimizer!(opt, localopt)
+            NLopt.maxeval!(opt, nstart * maxeval)
```
Vaibhavdixit02, line 94: Only these would need to be inside the conditional and the repeated code can be avoided

Vaibhavdixit02:
Oh lol Chris merged it already nvm

ChrisRackauckas:
That's fine. I just wanted to make sure @AStupidBear would get what they need for their work. But in the longer term, this is getting replaced with GalacticOptim.jl, so we can make sure do it right there. Hopefully that's up and running in a month or so, in which case all of our `sciml_train` examples will basically use that and it'll be a superset of functionality that makes everyone happy.

PR #235, opened 2020-05-12
Gigantic documentation update 2

ChrisRackauckas:


ChrisRackauckas:
@Emmanuel-R8 I'm done with what I was planning to write now. There's probably tons left to do, but at least every major point is written down except minibatching.

PR #236, opened 2020-05-13
Outing of constructors + Docstrings

Emmanuel-R8:
After nuking all my changes because I hate git rebase, here is a commit that "outs" the constructors defined in neural_de.jl. Plus a few minor changes to the docstrings. 

Beware: I cannot make it work the docstring generation on my machine. It seems that Documentor cannot find the source files although the path is pushed onto LOAD_PATH. If it works for any of you, then I'll welcome any piece of advice. Otherwise I'll keep debugging.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,4 +1,5 @@
-using Documenter, DiffEqFlux
+push!(LOAD_PATH, "../src/")
+using Documenter, DocStringExtensions, DiffEqFlux
 
 makedocs(
     sitename = "DiffEqFlux.jl",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,4 +1,4 @@
-using Documenter, DocStringExtensions, DiffEqFlux
+using Documenter, DiffEqFlux
 
 makedocs(
     sitename = "DiffEqFlux.jl",
--- a/src/neural_de.jl
+++ b/src/neural_de.jl
@@ -13,11 +13,11 @@ struct NeuralODE{M,P,RE,T,S,A,K} <: NeuralDELayer
 end
 
 """
-Constructs a continuous-time recurrant neural network, also known as a neural
-ordinary differential equation (neural ODE), with a fast gradient calculation
-via adjoints [1]. At a high level this corresponds to solving the forward
-differential equation, using a second differential equation that propagates the
-derivatives of the loss backwards in time.
+    Constructs a continuous-time recurrant neural network, also known as a neural
+    ordinary differential equation (neural ODE), with a fast gradient calculation
+    via adjoints [1]. At a high level this corresponds to solving the forward
+    differential equation, using a second differential equation that propagates the
+    derivatives of the loss backwards in time.
 
 ```julia
 NeuralODE(model,tspan,alg=nothing,args...;kwargs...)
@@ -89,7 +89,7 @@ struct NeuralDSDE{M,P,RE,M2,RE2,T,S,A,K} <: NeuralDELayer
 end
 
 """
-Constructs a neural stochastic differential equation (neural SDE) with diagonal noise.
+    Constructs a neural stochastic differential equation (neural SDE) with diagonal noise.
 
 ```julia
 NeuralDSDE(model1,model2,tspan,alg=nothing,args...;
@@ -162,7 +162,7 @@ struct NeuralSDE{P,M,RE,M2,RE2,T,S,A,K} <: NeuralDELayer
 end
 
 """
-Constructs a neural stochastic differential equation (neural SDE).
+    Constructs a neural stochastic differential equation (neural SDE).
 
 ```julia
 NeuralSDE(model1,model2,tspan,nbrown,alg=nothing,args...;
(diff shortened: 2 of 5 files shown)


docs/make.jl
```diff
@@ -1,4 +1,5 @@
-using Documenter, DiffEqFlux
+push!(LOAD_PATH, "../src/")
```
ChrisRackauckas, line 1:
```suggestion
```

src/neural_de.jl
```diff
@@ -34,29 +45,15 @@ Ref
...
-
-    function NeuralODE(model,tspan,solver=nothing,args...;kwargs...)
-        p,re = Flux.destructure(model)
-        new{typeof(model),typeof(p),typeof(re),
-            typeof(tspan),typeof(solver),typeof(args),typeof(kwargs)}(
-            model,p,re,tspan,solver,args,kwargs)
-    end
+function NeuralODE(model,tspan,solver=nothing,args...;kwargs...)
+    p,re = Flux.destructure(model)
+    NeuralODE(model,p,re,tspan,solver,args;kwargs)
```
ChrisRackauckas, line 50:
```suggestion
    NeuralODE(model,p,re,tspan,solver,args,kwargs)
```

but won't this infinite loop? I think the infinite looping due to the ambiguity on the splatted args is probably the reason they were made into inner constructors.

ChrisRackauckas:
You shouldn't have to do anything with LOAD_PATH and all of that jazz. If you have the module installed it should just work.

src/neural_de.jl
Emmanuel-R8: I made that change because I was getting confused how the parser would work since both args and kwargs are splatted.

docs/make.jl
Emmanuel-R8: I already added it.

Emmanuel-R8: I also added DocStringExtensions to have automatic generation of parts of the docstring. I won't use until I can resolve the current situation.

Emmanuel-R8:
I know. I have been banging my head on that. Trying anything I can think of. 

The whole thing makes no sense. For example. in the generated doc string of NeuralODE I see that "Arguments ≡≡≡≡≡≡≡≡" which has been deleted from the repo a while ago.

Clearing caches doesn't help. Chromium/Firefox give both show the same.

src/neural_de.jl
Emmanuel-R8: This works:

```
function testArgs(args...; kwargs...)
  @show args
  @show kwargs
end

testArgs(1, 2, 3; a = 4, b = 5, c = 6)

args = (1, 2, 3)
kwargs = Base.Iterators.Pairs(:a => 4,:b => 5,:c => 6)
```

This fails with `syntax: invalid "..." on non-final argument`

```
function testArgs(args..., kwargs...)
  @show args
  @show kwargs
end
```

ChrisRackauckas:
I explained why the change to outers is going to fail, so you might want to revert that. Without the package not being able to compile is then what's making it not possible to use the docstrings.

src/neural_de.jl
Emmanuel-R8: Or you are referring to the fact that there is already a keyword before for `solver`?

Then would that signature work:
`model::FastChain, tspan, args...; solver = nothing, kwargs...`

In which case, the first 3 work, the last one throws the same error.
```
function testArgs(args; a = 5, kwargs...)
  @show args
  @show a
  @show kwargs
end
testArgs(1; a = 3, b = 5, c = 6)

function testArgs(args, a = 5, kwargs...)
  @show args
  @show a
  @show kwargs
end
testArgs(1, a = 3, b = 5, c = 6)

function testArgs(args...; a = 5, kwargs...)
  @show args
  @show a
  @show kwargs
end
testArgs(1, 2; a = 3, b = 5, c = 6)

function testArgs(args..., a = 5, kwargs...)
  @show args
  @show a
  @show kwargs
end
testArgs(1, 2; a = 3, b = 5, c = 6)
```

Emmanuel-R8:
I don't understand where the infinite loop arises and why the inner constructor is needed. 

This is not to say I will not revert the change. Just looking for pointers to better understand.

ChrisRackauckas:
After you fix the kwargs, i.e. `,kwargs`, then it'll have the right number of arguments to call the outer constructor, causing an infinite loop.

Emmanuel-R8:
I get it now. You changed the call to the struct constructor, whereas I thought you changed the `;` in the call to the outer function. Apologies for creating the confusion.

ChrisRackauckas:
I really don't understand the point of these changes. To be blunt, it's not making the code nicer than just making one an inner constructor, i.e. https://github.com/SciML/DiffEqFlux.jl/pull/245 . I don't want you to put all of this time in expecting for this to merge as I see a very low probability of that happening.

Emmanuel-R8:
Chris - I see some recent changes that create a conflict with the pull request. 
Your changes simplify the call syntax. Do they also allow to create outer-constructors? 

My new version achieves the "outing". Should I resolve the conflicts to do the outing including your changes?

ChrisRackauckas:
I don't see a reason to do any outing. They could be made compatible with it, but I am not sure why. More important is the fix to Functor behavior.

Emmanuel-R8:
OK if I close this pull request?

ChrisRackauckas:
Yes

PR #241, opened 2020-05-16
Added example supervised NN-ODE

caseykneale:
So, in route to exploring an older interest I've had I started trying to brush up on the new API syntax for neural ODE's and stuff. Polished up an old example written by @jessebett , and it was suggested I make a PR for it :).

This PR has an example of an MNIST classifier that has an intermediate neural ODE layer in the model topology. The code was originally written by @jessebett  on his fork of the model zoo, but I updated it to match the new API.

This may not be "complete" enough for a tutorial, but it should get the ball rolling otherwise!


ChrisRackauckas:
Incorporated into https://github.com/SciML/DiffEqFlux.jl/pull/242, thanks!

PR #245, opened 2020-05-17
make all inner constructors for consistency

ChrisRackauckas:


ChrisRackauckas:
It's not redundant because parameters are only ever created once. We always need to get the restructure function though.

PR #248, opened 2020-05-18
missing global in docs example

dehann:


ChrisRackauckas:
Thanks! We should probably remove that list though, or actually make the animation again.

Issue #48, opened 2019-04-22
Using reverse-mode autodiff with a second order ODE problem


jgreener64:
So I was trying reverse-mode autodiff with a second order ODE problem. Here's what I have so far:

```julia
using DifferentialEquations
using Flux
using DiffEqFlux

u0 = Float32[0.; 2.]
du0 = Float32[0.; 0.]
tspan = (0.0f0, 1.0f0)
t = range(tspan[1], tspan[2], length=20)
p = param(Float32[])

model = Chain(Dense(2, 50, tanh), Dense(50, 2))

function f(du::TrackedArray, u::TrackedArray, p, t)
    model(u)
end
function f(du::AbstractArray, u::AbstractArray, p, t)
    Flux.data(model(u))
end

prob = SecondOrderODEProblem{false}(f, du0, u0, tspan, p)

function predict_rd()
    Flux.Tracker.collect(diffeq_rd(p, prob, Tsit5(), u0=(param(du0), param(u0)), saveat=t))
end

correct_pos = Float32.(transpose(hcat(collect(0:0.05:1)[2:end], collect(2:-0.05:1)[2:end])))

loss_n_ode() = sum(abs2, correct_pos .- predict_rd()[1:2, :])

data = Iterators.repeated((), 1000)
opt = ADAM(0.1)

cb = function ()
    println(loss_n_ode())
    loss_n_ode() < 0.01 && Flux.stop()
end

Flux.train!(loss_n_ode, params(model), data, opt, cb=cb)
```

This fails to run and the problem seems to be the `u0=(param(du0), param(u0))` argument to `diffeq_rd`. If I omit this argument or use `u0=(du0, u0)` then the initial conditions of the ODE problem, i.e. the untracked versions, are used. This means the ODE can be solved but the training doesn't change the neural network parameters.
Any ideas of how I can get round this, or am I doing something else wrong?

jgreener64:
Nice one.

PR #254, opened 2020-05-20
Second Order Neural ODEs

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqFlux.jl/issues/48


ChrisRackauckas:
Requires https://github.com/SciML/DiffEqSensitivity.jl/pull/243

Issue #244, opened 2020-05-17
Flesh out documentation


ChrisRackauckas:
There's a lot of parts of the documentation that can be better described:

- [ ] MNIST example can get a lot more text on what it's doing. Especially how it's performing minibatching.
- [ ] Every tutorial should have pictures. Before and after training pictures kind of thing.
- [ ] Benchmarks can get a lot more in there.
- [ ] We should have written down LaTeX forms of the equations we're fitting and have pictures of neural networks mixed in for each of the examples.
- [ ] It would be nice to get some animations in there.
- [ ] How mass matrix ODEs impose conditions could get a lot more detail.
- [ ] An example from a Hamiltonian or a graph network, i.e. https://github.com/SciML/DiffEqFlux.jl/issues/80
- [ ] It would be nice to get a PDE example in there, maybe just copying the one from the UDE stuff.
- [ ] Make sure every tutorial is covered by tests of some sort.
- [ ] Clean up minibatching docs: https://github.com/SciML/DiffEqFlux.jl/pull/267

JeremyFongSP:
For the Benchmarks, do you mean add more benchmarks or more text describing how it works/what it is doing?
For the mass matrix ODE, is this LL-NN-Stiff.md?

ChrisRackauckas:
>For the Benchmarks, do you mean add more benchmarks or more text describing how it works/what it is doing?

Yeah, some more benchmarks between adjoints, maybe getting a copy of some of the tutorials running in torchdiffeq to see the timing difference.

ChrisRackauckas:
>For the mass matrix ODE, is this LL-NN-Stiff.md?

yes

ChrisRackauckas:
Would be good to add a PDE-constrained optimized example:

```julia
using DelimitedFiles,Plots
using DiffEqSensitivity, OrdinaryDiffEq, Zygote, Flux, DiffEqFlux, Optim

# Problem setup parameters:
Lx = 10.0
x  = 0.0:0.01:Lx
dx = x[2] - x[1]
Nx = size(x)

u0 = exp.(-(x.-3.0).^2) # I.C

## Problem Parameters
p        = [1.0,1.0]    # True solution parameters
xtrs     = [dx,Nx]      # Extra parameters
dt       = 0.40*dx^2    # CFL condition
t0, tMax = 0.0 ,1000*dt
tspan    = (t0,tMax)
t        = t0:dt:tMax;

## Definition of Auxiliary functions
[...]

PR #260, opened 2020-05-23
Docs fix: remove parameter `p` from call of `predict_rd`

metanoid:
I couldn't get this example to run without this change.
What's still not clear to me (maybe the docs could explain this?) is why the  `loss_rd()` function isn't called like `loss_rd(p)`.
Surely `Flux.train` is trying to evaluate the impact of calling the `loss_rd` function at different values of `p`, so why is `p` not an argument to the function?

### DIFF
--- a/docs/src/examples/LV-Flux.md
+++ b/docs/src/examples/LV-Flux.md
@@ -20,7 +20,7 @@ prob = ODEProblem(lotka_volterra,u0,(0.0,10.0),p)
 Then we define our loss function with `concrete_solve` for the adjoint method:
 
 ```julia
-function predict_rd()
+function predict_rd(p)
   Array(concrete_solve(prob,Tsit5(),u0,saveat=0.1,reltol=1e-4))
 end
 loss_rd() = sum(abs2,x-1 for x in predict_rd(p))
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Flux uses implicit globals, so the right fix is actually to do `loss_rd() = sum(abs2,x-1 for x in predict_rd())`

metanoid:
Ah, ok, I hadn't realised that.  
The `sciml_train` function  seems to work differently, in its help:

```
sciml_train(loss, θ, opt, data = DEFAULT_DATA; cb, maxiters)
Optimizes the `loss(θ,curdata...)` function with respect to the parameter vector
`θ` iterating over the `data`.
```
From this it seems that for `sciml_train` you write the loss functions to expect to be passed the parameter vector to optimise, in contrast to `Flux.train!`.  Would it be worthwhile to add a section to the docs pointing out such differences?  If they are already there, I missed them.

ChrisRackauckas:
>From this it seems that for sciml_train you write the loss functions to expect to be passed the parameter vector to optimise, in contrast to Flux.train!.

Indeed that is the case. The two forms are quite different, and have different optimizations available. 
That's why we have different tutorials for the Flux parts, and stuff like https://diffeqflux.sciml.ai/dev/Flux/ . If there's a way to rearrange the discussion of this difference to make it easier, we can definitely change it up.

ChrisRackauckas:
Thanks! For more documentation discussion, we can take it to https://github.com/SciML/DiffEqFlux.jl/issues/244

PR #261, opened 2020-05-23
Update LV-delay.md to include training step

metanoid:
For consistency with the other examples, this example needs to finish off with actually calling the `sciml_train` function as a demonstration


ChrisRackauckas:
Thanks!

PR #266, opened 2020-05-25
[WIP] Add Documentation to LV-NN-Stiff.md

JeremyFongSP:
Adding Documentation in the same style as Supervised-NN-ODE-MNIST.
This may take a little while, I have minimal experience with stiff equations. Walkthrough/Comments/Suggestions greatly appreciated.

### DIFF
--- a/docs/src/examples/LV-NN-Stiff.md
+++ b/docs/src/examples/LV-NN-Stiff.md
@@ -101,14 +101,11 @@ tspan = (0.0,1.0)
 p = [0.04, 3e7, 1e4]
 ```
 
-`u₀` = Initial Conditions
-
-`M` = Semi-explicit Mass Matrix (last row is the constraint equation and are therefore
+- `u₀` = Initial Conditions
+- `M` = Semi-explicit Mass Matrix (last row is the constraint equation and are therefore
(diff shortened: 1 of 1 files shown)


docs/src/examples/LV-NN-Stiff.md
```diff
@@ -96,18 +101,35 @@ tspan = (0.0,1.0)
...
 ```
 
+`u₀` = Initial Conditions
+
+`M` = Semi-explicit Mass Matrix (last row is the constraint equation and are therefore
+all zeros)
+
+`tspan` = Time span over which to evaluate
+
+`p` = parameters `k1`, `k2` and `k3` of the differential equation above
```
ChrisRackauckas, line 111:
```suggestion
- `u₀` = Initial Conditions
- `M` = Semi-explicit Mass Matrix (last row is the constraint equation and are therefore
all zeros)
- `tspan` = Time span over which to evaluate
- `p` = parameters `k1`, `k2` and `k3` of the differential equation above
```

ChrisRackauckas:
No longer WIP?

JeremyFongSP:
> No longer WIP?
I think it's done

PR #267, opened 2020-05-26
Add mini batch docs

collinwarner:


docs/src/examples/minibatch.md
```diff
@@ -0,0 +1,81 @@
+# Training a Neural Ordinary Differntial Equation with Mini-Batching
+
+When training universal differential equations it is often helpful to batch our data. This is particularly useful when working with large sets of training data. Let us take a look at how this works with the Lotka-Volterra equation.
```
ChrisRackauckas, line 3: we're moving towards a style where there's a copy-pastable code at the top, and then a version with interspersed explanation below it.

docs/src/examples/minibatch.md
```diff
@@ -0,0 +1,81 @@
...
+```julia
+dudt2 = FastChain((x,p) -> x.^3,
+            FastDense(2,50,tanh),
+            FastDense(50,2))
+n_ode = NeuralODE(dudt2,tspan,Tsit5(),saveat=t)
+```
+
+In our model we used the `x -> x.^3` assumption in the model. By incorporating structure into our equations, we can reduce the required size and training time for the neural network, but a good guess needs to be known!
+
+From here we build a loss function around it. The `NeuralODE` has an optional second argument for new parameters which we will use to iteratively change the neural network in our training loop. We will use the network's output against the time series data. To add support for batches of size `k` we need to add parameters representing the start index and size of our batch to our loss function:
```
ChrisRackauckas, line 36: this seems the same as the neural ODE tutorial?

docs/src/examples/minibatch.md
ChrisRackauckas, line 3: explain what minibatching is, why it's used, and generally how it's done.

docs/src/examples/minibatch.md
```diff
@@ -0,0 +1,81 @@
...
+  end
+  
+```
+
+and then create a generator, that produces `MAX_BATCHES` random tuples `(start index, batch size)`:
+
+```julia
+  MAX_BATCHES = 1000
+  k = 15 #batch size
+  data = ((rand(1:size(ode_data)[2] -k), k) for i in 1:MAX_BATCHES)
```
ChrisRackauckas, line 56: demonstrate how this is minibatching by showing a for loop with `@show` in it. Point out other tools for minibatching like Flux's DataLoader and stuff in JuliaML (MLDataUtils.jl I think has something?)

PR #286, opened 2020-06-09
Fix for RecursiveArrayTools broadcast fix

ChrisRackauckas:


YingboMa:
Umm, this looks a little weird. Is it a bug in RecursiveArrayTools?

ChrisRackauckas:
Nah it makes sense. The broadcast between two VectorOfArrays produces a VectorOfArray, so the broadcast between two solutions does the same. It's just not what we wanted here: we wanted an array, so it requires choosing to make something an array.

Issue #282, opened 2020-06-08
Library-tized FFJORD


d-netto:
@ChrisRackauckas: base distributions and the method of calculation of the trace of the Jacobian can be changed by the user.
```Julia

using Distributions
using DifferentialEquations
using Flux, DiffEqFlux
using Zygote, Tracker
using Plots

##
##Library-tized version
##

abstract type CNFLayer <: Function end
Flux.trainable(m::CNFLayer) = (m.p,)

struct FFJORD{M,P,RE,ContinuousUnivariateDistribution,Bool,T,A,K} <: CNFLayer
    model::M
    p::P
    re::RE
    baseDist::ContinuousUnivariateDistribution
    MonteCarlo::Bool
    tspan::T
    args::A
    kwargs::K

    function FFJORD(model,tspan,args...;p = nothing,baseDist=nothing,MonteCarlo=nothing,kwargs...)
        _p,re = Flux.destructure(model)
        if p === nothing
            p = _p
        end
        if baseDist === nothing
            baseDist = Normal(0,1)
        end
        if MonteCarlo === nothing
            MonteCarlo = false
        end
        new{typeof(model),typeof(p),typeof(re),typeof(baseDist),
            typeof(MonteCarlo),typeof(tspan),typeof(args),typeof(kwargs)}(
            model,p,re,baseDist,MonteCarlo,tspan,args,kwargs)
    end
end

function ffjord(du,u,p,t,re,MonteCarlo,e)
    z = @view u[1:end-1]
    m = re(p)
    fz, back = Zygote.pullback(m,z')
    if MonteCarlo
        eJ = back(e)[1]
        jac = -(eJ.*e)[1]
    else
        jac = -sum(back(1)[1])
    end
    du[1:end-1] .= m(z)
    du[end] = jac
end

function (n::FFJORD)(x,p=n.p,MonteCarlo=n.MonteCarlo)
    e = MonteCarlo ? randn(Float32,size(p)[1]) : nothing
    ffjord_ = (du,u,p,t)->ffjord(du,u,p,t,n.re,MonteCarlo,e)
    prob = ODEProblem{true}(ffjord_,nothing,n.tspan,nothing)
    pred = concrete_solve(prob,Tsit5(),[x,0f0],p,n.args...;n.kwargs...)[:,end]
    pz = n.baseDist
    z = pred[1]
    delta_logp = pred[2]
    logpz = logpdf(pz, z)
    logpx = logpz - delta_logp
[...]

Issue #47, opened 2019-04-18
FFJORD


ChrisRackauckas:
```julia
using OrdinaryDiffEq
using Distributions
using Flux, DiffEqFlux, Tracker

# Neural Network
nn = Chain(Dense(1,1,tanh))
p = Tracker.data(DiffEqFlux.destructure(nn))
DiffEqFlux.restructure(nn,p)([1.0])
tspan = Float32.((0.0, 10.0))
function cnf(du,u,p,t)
  z = @view u[1:end-1]
  m = DiffEqFlux.restructure(nn,p)
  du[1:end-1] = m(z)
  fz, back = Tracker.forward(z->m(z),z')
  e = randn(size(z)[1])
  eJ = back(e)[1]
  eJe = Tracker.data((eJ .* e)[1])
  du[end] = -eJe
end

Tracker.collectmemaybe(z)

prob = ODEProblem(cnf,nothing,tspan,nothing)

p = param(Float32[0.0, 0.0]) # Initial Parameter Vector
params = Params([p])

function predict_adjoint(x)
    diffeq_adjoint(p,prob,Tsit5(),u0=[x;false],
                   saveat=0.0:0.1:10.0,
                   sensealg=DiffEqFlux.SensitivityAlg(quad=false,
                                backsolve=true,autojacvec=true))
end

function loss_adjoint(xs)
    pz = Normal(0.0, 1.0)
    preds = [predict_adjoint(x)[:,end] for x in xs]
    z = [pred[1] for pred in preds] # TODO better slicing
    delta_logp = [pred[2] for pred in preds]

    logpz = logpdf.(pz, z)
    logpx = logpz - delta_logp
    loss = -mean(logpx)
end

opt = ADAM(0.1)

raw_data = [Float32[rand(Normal(2.0, 0.1)) for i in 1:100]]
data = Iterators.repeated(raw_data, 100);

Flux.train!(loss_adjoint, params, data, opt)

# check whether it looks standard normal
using Plots

preds = [predict_adjoint(r)[:,end] for r in raw_data[1]];
```

ChrisRackauckas:
```julia
using OrdinaryDiffEq
using Distributions
using Flux, DiffEqFlux, Tracker

# Neural Network
nn = Dense(1,1,tanh)
p = Tracker.data(DiffEqFlux.destructure(nn))
tspan = Float32.((0.0, 10.0))
function ffjord(du,u,p,t)
  z = @view u[1:end-1]
[...]

PR #287, opened 2020-06-09
FFJORD

d-netto:
Closes https://github.com/SciML/DiffEqFlux.jl/issues/47 and closes https://github.com/SciML/DiffEqFlux.jl/issues/282


ChrisRackauckas:
Needs a test

src/ffjord.jl
```diff
@@ -0,0 +1,56 @@
...
+    model::M
+    p::P
+    re::RE
+    baseDist::ContinuousUnivariateDistribution
+    MonteCarlo::Bool
+    tspan::T
+    args::A
+    kwargs::K
+
+    function FFJORD(model,tspan,args...;p = nothing,baseDist=nothing,MonteCarlo=nothing,kwargs...)
```
ChrisRackauckas, line 14:
```suggestion
    function FFJORD(model,tspan,args...;p = nothing,basedist=Normal(0,1),monte_carlo=false,kwargs...)
```

Following the standard naming conventions for Julia variables

src/ffjord.jl
```diff
@@ -0,0 +1,56 @@
...
+        _p,re = Flux.destructure(model)
+        if p === nothing
+            p = _p
+        end
+        if baseDist === nothing
+            baseDist = Normal(0,1)
+        end
+        if MonteCarlo === nothing
+            MonteCarlo = false
+        end
```
ChrisRackauckas, line 24:
```suggestion
```

There's no need to treat these in a separate manner

ChrisRackauckas:
@jessebett

src/ffjord.jl
d-netto: Ok. Should I also change the fields in the struct to follow the convention?

src/ffjord.jl
d-netto: What I had in mind is that the user could choose, for example, to use the default base distribution and set monte_carlo=true , to use an alternative distribution and the default monte_carlo, etc. 

monte_carlo and basest would be independent in this sense.

d-netto:
@ChrisRackauckas What do you think about changing basedist to also support multivariate distributions?

src/ffjord.jl
ChrisRackauckas: no, normal keyword argument handling works just fine in that case. You don't need `=== nothing` and then special handling.

src/ffjord.jl
ChrisRackauckas: yes

src/ffjord.jl
ChrisRackauckas: see that https://github.com/SciML/DiffEqFlux.jl/pull/287#discussion_r437083643 sets those values as the default. `p` is handled differently because `destructure` is where `p` comes from, unless the user passed in `p`: that situation is complex enough to not just be a keyword argument.

d-netto: oh, didn't see that it was changed to false

ChrisRackauckas:
The test needs to get enabled by adding it to runtests.jl

d-netto:
@ChrisRackauckas test throws an error because Distributions is not in current path.

ChrisRackauckas:
Add Distributions as a test dependency: https://github.com/SciML/DiffEqFlux.jl/blob/master/Project.toml#L44-L56

docs/Project.toml
```diff
@@ -1,5 +1,57 @@
...
+DelayDiffEq = "bcd4f6db-9728-5f36-b5f7-82caef46ccdb"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+MultistartOptimization = "3933049c-43be-478e-a8bb-6e0f7fd53575"
+NLopt = "76087f3c-5699-56af-9a33-bf431cd00edd"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+StochasticDiffEq = "789caeaf-c7a9-5a7d-9973-96adeb23e2a0"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
```
ChrisRackauckas, line 54: I think you accidentally changed the Project.toml in the docs

d-netto: Yeah. Dumb question: should I keep the Project.toml as it's in my repo and add the dependencies there or I should copy paste the Project.toml in DiffEqFlux and add some lines to include Distributions.jl?

ChrisRackauckas: revert this and change the one in the top level.

PR #302, opened 2020-06-10
FFJORD for multivariate distributions

d-netto:


ChrisRackauckas:
unrelated stochastic test failure

PR #303, opened 2020-06-11
Update Neural ODE MNIST Example

avik-pal:
This PR updates the Supervised Neural ODE MNIST examples to include the following:
1. Use the newly added `Flux.Data.DataLoader` for automatic batching and shuffling the data
2. The previous example only reported the training accuracy which is considered bad practice so it has been modified to report both train and test accuracies.

### DIFF
--- a/docs/src/examples/Supervised-NN-ODE-MNIST.md
+++ b/docs/src/examples/Supervised-NN-ODE-MNIST.md
@@ -6,99 +6,115 @@ on **GPUs** with **Minibatching**.
 (Step-by-step description below)
 
 ```julia
-using DiffEqFlux, OrdinaryDiffEq, Flux, MLDataUtils, NNlib
+using DiffEqFlux, OrdinaryDiffEq, Flux, NNlib, MLDataUtils, Printf
 using Flux: logitcrossentropy
-using MLDatasets: MNIST
-using CuArrays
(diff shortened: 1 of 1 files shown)


docs/src/examples/Supervised-NN-ODE-MNIST.md
```diff
@@ -134,30 +151,40 @@ one hot encoding (`LabelEnc.OneOfK`).
...
+    # Load MNIST
+    imgs, labels_raw = MNIST.traindata();
+    # Process images into (H,W,C,BS) batches
+    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
+    y_data = onehot(labels_raw)
+    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
+                                                         p = train_split)
+    return (
+        # Use Flux's DataLoader to automatically minibatch and shuffle the data
+        DataLoader(gpu.(collect.([x_train, y_train]))...; batchsize = batchsize,
```
ChrisRackauckas, line 173: Is the splat needed here?

avik-pal: Fixed it in the latest commit

PR #304, opened 2020-06-11
Added example with ComponentArrays

jonniedie:
Example to show building neural layers from scratch using ComponentArrays.jl.


ChrisRackauckas:
Fantastic! We should probably start moving more in this direction.

PR #315, opened 2020-06-17
Use Float32s in ComponentArray example

jonniedie:
These should be `ComponentArray{Float32}`s to match the other examples.

### DIFF
--- a/docs/src/Flux.md
+++ b/docs/src/Flux.md
@@ -167,7 +167,7 @@ ode_data = Array(solve(prob, Tsit5(), saveat = t))
 
 Now, we'll make a function that creates dense neural layer components. It is similar to `Flux.Dense`, except it doesn't handle the activation function. We'll do that separately.
 ```julia
-dense_layer(in, out) = ComponentArray(W=glorot_uniform(out, in), b=zeros(out))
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
That's nice syntax.

jonniedie:
The syntax works for inner array types too (as long as they have a compatible constructor). I just haven't documented that yet because I don't quite have the broadcasting story down quite right for them.
```julia
julia> using CuArrays

julia> x = ComponentArray{CuArray{Float32}}(a=rand(2,2), b=(a=rand(), b=rand()))
ComponentVector{CuArray{Float32,1,Nothing}}(a = Float32[0.2098118 0.16323364; 0.41578043 0.13620397], b = (a = 0.9387826, b = 0.9181254))

julia> x.a*x.b
2-element CuArray{Float32,1,Nothing}:
 0.3468366
 0.5153797
```
That works, but this doesn't (but it would work for things that fall back to normal `Array`s like `StaticArray`s do):
```julia
julia> x.a .+ x.b
ERROR: MethodError: no method matching +(::Float32, ::SubArray{Float32,0,CuArray{Float32,1,CuArray{Float32,1,Nothing}},Tuple{Int64},true})
...
```
Broadcasting in general has taken some effort. I should have it down pretty soon though.

PR #316, opened 2020-06-17
Augmented Neural DE

### LINKED ISSUE #288: Augmented neural ODEs
https://arxiv.org/abs/1904.01681

avik-pal:
Fixes #288.

TODOs before it is ready for review:
- [x] Minimal Example of Augmented ODE
- [x] Test working with the other DE layers
- [x] Documentation
- [x] Unit Tests
- [x] More details in the example


docs/src/examples/Augmented-NN-ODE.md
```diff
@@ -0,0 +1,250 @@
...
+
+# Expected Output
+
+```julia
+┌ Info: Generating Dataset
+└ @ Main In[11]:60
+┌ Info: Training Neural ODE
+└ @ Main In[11]:75
+┌ Info: Epoch 1
+└ @ Main /home/avik-pal/.julia/packages/Flux/Fj3bt/src/optimise/train.jl:121
```
ChrisRackauckas, line 110: maybe we don't want to do it as `@info` to get rid of the line number?

avik-pal:
I added some tests, but not sure why the CI is not triggered.

test/runtests.jl
```diff
@@ -24,6 +24,7 @@ end
 
 if GROUP == "All" || GROUP == "DiffEqFlux" || GROUP == "BasicNeuralDE"
     @safetestset "Neural DE Tests" begin include("neural_de.jl") end
+    @safetestset "Augmented Neural DE Tests" begin include("fast_neural_ode.jl") end
```
Vaibhavdixit02, line 27: Looks like a wrong file is used for the test here?

src/neural_de.jl
```diff
@@ -470,3 +470,37 @@ function (n::NeuralODEMM{M})(x,p=n.p) where {M<:FastChain}
...
+    nde::DE
+    adim::Int
+end
+
+Flux.@functor AugmentedNDELayer (nde,)
+
+(ande::AugmentedNDELayer)(x, args...) = ande.nde(augment(x, ande.adim), args...)
+
+augment(x::AbstractVector{S}, augment_dim::Int) where S =
+    cat(x, zeros(S, (augment_dim,)), dims = 1)
```
ChrisRackauckas, line 500: I see, yes because the augmented dimensions have null dynamics this works.

avik-pal:
The test failures seem to be from something unrelated to the Augmentation Layer

PR #317, opened 2020-06-18
rename docs/src/examples markdown files

JeremyFongSP:


ChrisRackauckas:
make.jl also needs an update

PR #323, opened 2020-06-21
update requires for CUDA.jl

ChrisRackauckas:
Currently not possible to use because of Adapt.jl, but this should be right?


ChrisRackauckas:
@maleadt is this right?

maleadt:
Yeah that looks OK.

PR #325, opened 2020-06-22
Tensor Layer: documentation and optimization

d-netto:


ChrisRackauckas:
We should get a doc page where all of these layer docstrings go, with a description on the top for how to put them all together and what they are for.

d-netto:
> We should get a doc page where all of these layer docstrings go, with a description on the top for how to put them all together and what they are for.

Yup, in progress.

src/tensor_product_layer.jl
```diff
@@ -9,7 +9,7 @@ struct TensorLayer{M<:Array{TensorProductBasis},P<:AbstractArray,Int} <: Abstrac
         for basis in model
             number_of_weights *= basis.n
         end
-        p = rand(out*number_of_weights)
+        p = zeros(out*number_of_weights)
```
ChrisRackauckas, line 12:
```suggestion
        p = randn(out*number_of_weights)
```

Having them all positive is hard to fit, having them all negative is hard to fit, having them all zero makes it very poorly conditioned, so normally Wishart or normally distributed random numbers are effective.

docs/make.jl
```diff
@@ -32,13 +32,15 @@ makedocs(
...
+        "Basis" => "basis_docs.md",
         "Continuous Normalizing Flows Layer" => "CNFLayer.md",
         "Neural Differential Equation Layers" => "NeuralDELayers.md",
         "Use with Flux Chain and train!" => "Flux.md",
         "FastChain" => "FastChain.md",
         "GPUs" => "GPUs.md",
         "sciml_train" => "Scimltrain.md",
-        "Benchmark" => "Benchmark.md"
+        "Benchmark" => "Benchmark.md",
+        "Tensor Product Layer" => "tensor_layer.md"
```
ChrisRackauckas, line 43: Tensor product layer should probably go with basis

Issue #327, opened 2020-06-25
Neural ODE trains but throws an error when trying to show the result from Optim


John-Boik:
This is a followup to the closed issue [243](https://github.com/SciML/DiffEqFlux.jl/issues/243) of nearly the same name. I also cannot run script [Neural-ODE-Example-1](https://diffeqflux.sciml.ai/dev/GPUs/#Neural-ODE-Example-1) to completion. 

This is for the script that starts:
```
using DiffEqFlux, OrdinaryDiffEq, Flux, Optim, Plots, CuArrays, DiffEqSensitivity
CuArrays.allowscalar(false) # Makes sure no slow operations are occuring
```

After reaching the max number of iterations, I get the message:
```
 * Status: failure (reached maximum number of iterations)

 * Candidate solution
Error showing value of type Optim.MultivariateOptimizationResults{ADAM,Float64,CuArray{Float32,1,Nothing},Float64,Float32,Nothing,Bool}:
ERROR: scalar getindex is disallowed
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] assertscalar(::String) at /home/jboik/.julia/packages/GPUArrays/JqOUg/src/host/indexing.jl:41
 [3] getindex(::CuArray{Float32,1,Nothing}, ::Int64) at /home/jboik/.julia/packages/GPUArrays/JqOUg/src/host/indexing.jl:96
 [4] iterate at ./abstractarray.jl:913 [inlined]
 [5] iterate at ./abstractarray.jl:911 [inlined]
 [6] iterate at ./iterators.jl:641 [inlined]
 [7] iterate at ./iterators.jl:639 [inlined]
 [8] iterate at ./generator.jl:44 [inlined]
 [9] collect(::Base.Generator{Base.Iterators.Take{CuArray{Float32,1,Nothing}},Optim.var"#2#4"}) at ./array.jl:665
 [10] show(::IOContext{REPL.Terminals.TTYTerminal}, ::Optim.MultivariateOptimizationResults{ADAM,Float64,CuArray{Float32,1,Nothing},Float64,Float32,Nothing,Bool}) at /home/jboik/.julia/packages/Optim/L5T76/src/types.jl:242
[...]

PR #332, opened 2020-06-27
fix output statuses

ChrisRackauckas:
This was bugging me. Fixes it so that way maxiters counts as success. Fixes https://github.com/SciML/DiffEqFlux.jl/issues/327


ChrisRackauckas:
@Vaibhavdixit02 @abhigupta768 let's make sure this happens to GalacticOptim.jl as well

PR #338, opened 2020-07-04
changes in tensor layer

d-netto:


ChrisRackauckas:
The test error looks real

d-netto:
Dimensions were inverted.

Project.toml
```diff
@@ -20,6 +22,7 @@ ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
Vaibhavdixit02: This shouldn't be added as a dependency.

d-netto: yeah, I believe it's fixed in the latest commit

Vaibhavdixit02: My bad, sorry

d-netto:
Tests passed. Travis outputting "in progress" for some reason.

ChrisRackauckas:
Great. I just realized... can we test whether this works with minibatching? I am not sure it does.

d-netto:
Ok. I'll create a test for mini-batching.

PR #339, opened 2020-07-06
Spline layer [WIP]

d-netto:


d-netto:
`DataInterpolations` seems to be mutating something and `Zygote` throws an error locally.

d-netto:
No support for splines because `DataInterpolations` is still mutating.

ChrisRackauckas:
Wow... that converges super fast. I'll get this merged, and the follow up could then just add tests with the cubic splines when they're ready.

Issue #80, opened 2019-10-02
Hamiltonian Graph Networks with ODE Integrators


ChrisRackauckas:
https://arxiv.org/pdf/1909.12790.pdf

Need a good name for it. Tracking it here, and we should make sure to have a good implementation of it that's easy to plug and play for "non-diffeq people".

yuehhua:
Count on me! :smile: 
Graph network is introduced by [Relational inductive biases, deep learning, and graph networks](https://arxiv.org/pdf/1806.01261.pdf) by Google. Graph network is composed of meta layer, which processes data in three layers in edge level, vertex level and global level, respectively. I will implement `Meta` layer in [GeometricFlux](https://github.com/yuehhua/GeometricFlux.jl), then integrate with ODE integrators.

yuehhua:
OK, I think I have time to work on it.
For meta layer, I will have `Meta` which is abstract type with three main abstract method `update_edge`, `update_vertex` and `update_global`.
If DiffEqFlux is loaded, `ContinuousMeta` is loaded, which is a subtype of `Meta` but still abstract.
I hope any DEs supposed in DiffEqFlux could be used under this framework.
Those update functions for graph are interfaces for users to implement their own neural network and plug in, so we have to put those update functions into a DE problem to make them work.
Time is also an important parameters to put in, but I still don't have idea how to passing these arguments.

yuehhua:
I prototyped the `Meta` layer in a good abstraction and tried to integrate with DiffEqFlux.
Then I realized that I don't have to do that!
Because GeometricFlux is an extension of Flux layer, it is already integrated by DiffEqFlux and it is fully composable!
All users need to do is replace the regular deep learning layers into GeometricFlux's layers!
Awesome Julia!
All I have to do is to make a example for it and maybe I will need some data.

yuehhua:
So far, `Meta` layer abstraction and message passing scheme is done in GeometricFlux. However, GeometricFlux utilize Flux with Zygote, so DiffEqFlux should also have Zygote support. Issues relate to #52 and JuliaDiffEq/DiffEqSensitivity.jl#71.

Zymrael:
https://arxiv.org/abs/1911.07532 "Graph Neural Ordinary Differential Equations" (GDE)

[...]

PR #340, opened 2020-07-08
Neural Graph Differential Equation Example

avik-pal:
This modifies the GCN demo in GeometricFlux to use a Neural GDE.

https://github.com/yuehhua/GeometricFlux.jl/pull/62 needs to be merged and a release has to be tagged before it works properly.


yuehhua:
Thank you for implement this example. I just got everything done at yuehhua/GeometricFlux.jl#64.
I will tag it as new release ASAP.
Related issue #80

yuehhua:
Wait for new release JuliaRegistries/General#17623

ChrisRackauckas:
Let's get a test added that will capture errors in the tutorial. That'll make it easier to maintain over time. Probably just a simple example where a NGDE is trained on some silly loss and just make sure the loss goes down.

yuehhua:
I just check travis-ci, and it shows the following words:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```
It seems that it failed because the build time is too long and timeout happens. Might not due to the code itself but I am not sure about the root cause. Whenever I found that the CI failed not because of my code, I just restart it and it get passed.

ChrisRackauckas:
Yeah, Julia v1.4 is the high of `using` times... at least v1.5 fixes it.

test/neural_gde.jl
```diff
@@ -0,0 +1,35 @@
+using OrdinaryDiffEq, Flux, Test, GeometricFlux
```
ChrisRackauckas, line 1:
```suggestion
using OrdinaryDiffEq, DiffEqFlux, Flux, Test, GeometricFlux
```

ChrisRackauckas:
Completed the test change in #355 and merged these commits. Thanks!

PR #356, opened 2020-07-11
Use travis_wait

christopher-dG:
If it still times out, we can pass a number of minutes to `travis_wait` i.e. `travis_wait 60 julia -e '...'`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ env:
   - GROUP=Integration
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
https://travis-ci.org/github/SciML/DiffEqFlux.jl/jobs/707259744

That doesn't look right.

christopher-dG:
I think it's working properly (just timing out still), IIRC that wait function dumps all the output at the end instead of streaming it.

ChrisRackauckas:
oh okay

PR #365, opened 2020-07-16
revised minibatch docs

collinwarner:
Changed to the format where the code is in pasteable form at the top. Implemented mini batching with Flux.dataLoader and eachbatch in MLDataUtils. Expanded upon mini batching explanation. Changed ODE used as a demo to netwons cooling.


docs/src/examples/minibatch.md
```diff
@@ -1,65 +1,155 @@
...
+tspan = (0.0f0, 1.5f0)
+
+t = range(tspan[1], tspan[2], length=datasize)
+true_prob = ODEProblem(true_sol, u0, tspan)
+ode_data = Array(solve(true_prob, Tsit5(), saveat=t))
+
 ```
 
-Now let's define a neural network with a `NeuralODE` layer. First we define the layer. Here we're going to use `FastChain`, which is a faster neural network structure for NeuralODEs:
+Now we define a neural-network using a linear approximation with 8 hidden layers.
```
ChrisRackauckas, line 100: one hidden layer of size 8

ChrisRackauckas, line 100:
```suggestion
Now we define a neural-network using a linear approximation with 1 hidden layer of 8 neurons.  
```

ChrisRackauckas:
Can you resolve the merge conflicts?

Issue #368, opened 2020-07-18
The MNIST Neural ODE Classifier tutorial is broken 


mbeltagy:
DiffEqFlux seems to be using a older version of Flux, though the code appears to written for Flux 0.11.0. 

The code in [example](https://diffeqflux.sciml.ai/dev/examples/mnist_neural_ode/) gives the following errors

```julia
julia> img, lab = train_dataloader.data[1][:, :, :, 1:1], train_dataloader.data[2][:, 1:1]

ERROR: BoundsError: attempt to access (Array[Float32[0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0]

Float32[0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0]

Float32[0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0]

...

Float32[0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0]

Float32[0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0]

Float32[0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0], [0 0 … 0 0; 0 0 … 1 0; … ; 0 0 … 0 1; 0 0 … 0 0]],)
  at index [2]
```

ChrisRackauckas:
We plan to update whenever GeometricFlux is compatible, so we should have a syntax that reflects that because I hope that happens today.

ChrisRackauckas:
Fix.

PR #369, opened 2020-07-18
MNIST example DataLoader syntax bugfix

mbeltagy:
Fixed the DataLoader syntax issue in https://github.com/SciML/DiffEqFlux.jl/issues/368


ChrisRackauckas:
@DhairyaLGandhi @avik-pal is this a better fix than the splat?

avik-pal:
Yes. The splatting syntax is (/will be) deprecated I believe.

DhairyaLGandhi:
Lgtm too, the splatting behavior should be gone soon.

PR #370, opened 2020-07-20
Add support for Hamiltonian Neural Networks

### LINKED ISSUE #53: HamiltonianProblem not supported
Hi, thanks for the awesome package! Unfortunately I'm having issues using it for my work. I basically want to integrate a set of equations of motions defined by a Hamiltonian `H(p, x)` that is too complex to differentiate by hand. The code below sets up the problem:
```julia
using DifferentialEquations
using DiffEqFlux
using Flux

function H(p, x, params)  # complicated (randomly selected) Hamiltonian
    return params[1] * (1 + x[1]^2)^(1/3) * (1 + p[1]^2)^(1/4)
end

function getprob(params)  # set up problem
    tspan = (0.0, 1.0)
    x0, p0 = [0.01], [0.05]
    return HamiltonianProblem(H, p0, x0, tspan, params)
end
```
Now I want to differentiate solutions of the equations of motion with respect to the parameter vector. The function below sets this up and tries to run the ODE solver:
```julia
function testsolve(p0=[0.1])
    prob = getprob(p0)  # set up problem

    p = param(p0)  # initialize parameter

    function predict_rd()
        Tracker.collect(diffeq_rd(p, prob, Tsit5(), saveat=0.1))
    end
    
    println("test: ", predict_rd())  # try running the diff eq solver
end
```
When I run `testsolve()` I get the error
```julia
ERROR: MethodError: *(::Tracker.TrackedReal{Float64}, ::ForwardDiff.Dual{ForwardDiff.Tag{DiffEqPhysics.PhysicsTag,Float64},Float64,1}) is ambiguous. Candidates:
  *(a::Tracker.TrackedReal, b::Real) in Tracker at /Users/acoogan/.julia/packages/Tracker/RRYy6/src/lib/real.jl:94
  *(x::Real, y::ForwardDiff.Dual{Ty,V,N} where N where V) where Ty in ForwardDiff at /Users/acoogan/.julia/packages/ForwardDiff/N0wMF/src/dual.jl:140
Possible fix, define
  *(::Tracker.TrackedReal, ::ForwardDiff.Dual{Ty,V,N} where N where V)
```
followed by a massive stack trace.

It seems like there's a conflict between the datatypes used to define the equations of motion and the parameters tracked by Flux. Is there a way to fix this?

I was not able to get this to work using Zygote rather than ForwardDiff to differentiate `H` -- hopefully I'm missing something about how to use nested autodifferentiation?

### LINKED ISSUE #80: Hamiltonian Graph Networks with ODE Integrators

### LINKED ISSUE #357: Hamiltonian Neural Networks
https://greydanus.github.io/2019/05/15/hamiltonian-nns/

avik-pal:
Fixes #53
Fixes #357
Fixes #80 when this PR is used together with #355 

TODOs:
- [x] Implement Neural Hamiltonian DE Layer
- [x] Example for using the layer
- [ ] ~Working on GPU?~ Will be fixed with Keno's AD https://github.com/SciML/DiffEqFlux.jl/pull/370#issuecomment-663846037
    - [ ] ~HNN layer~
    - [ ] ~Neural Hamiltonian DE Layer~
- [x] Documentation
- [x] Tests


src/neural_de.jl
```diff
@@ -502,3 +502,54 @@ augment(x::AbstractArray{S, T}, augment_dim::Int) where {S, T} =
...
+            model,p,re,tspan,args,kwargs)
+    end
+end
+
+function (n::NeuralHamiltonianDE)(x,p=n.p)
+    function neural_hamiltonian!(du,u,p,t)
+        H = Flux.gradient(u -> n.re(p)(u)[1], u)[1]
+        n = size(x)[1] ÷ 2
+        du[1:n] .= -H[1:n]
+        du[(n + 1):end] .= H[(n + 1):end]
```
avik-pal, line 538: I am not sure how to do it in a way that is GPU friendly. Any suggestions for improvement?

ChrisRackauckas: using Slices like that is okay, it's just two kernels. But you can make it one kernel by using a fill array with a `-1` and a 1, or just allocating the array of 1 and -1 (maybe caching it). That should be a relatively minor improvement though.

avik-pal:
Zygote seems to complain about mutating arrays when backpropagating through the gradient of the NN. Here's a small repro:

```julia
using Flux, Zygote

model = Chain(
    Dense(2, 10),
    Dense(10, 1)
)

forward(m, x) = gradient(x -> m(x)[1], x)[1]

x = rand(2, 1)

gradient(() -> sum(forward(model, x) - x), Flux.params(model))
```
Is this to be done in some other way?

ChrisRackauckas:
Nesting Zygote is an issue. I think you need to use ReverseDiff over Zygote, or Zygote over ReverseDiff.

avik-pal:
Yes, ReverseDiff over Zygote does solve this issue.

avik-pal:
For training HNNs the standard way ([torchdyn](https://torchdyn.readthedocs.io/en/latest/tutorials/08_hamiltonian_nets.html), [original paper](https://github.com/greydanus/hamiltonian-nn/blob/master/hnn.py)) seems to be to supervise on the gradients of the network, rather than on the solution from the ODE Solver. I was wondering if we should have a `HamiltonianNN` layer separate from the `NeuralHamiltonianDE` which simply returns the gradient and the latter just acts as a wrapper over the former to compute the solutions of the ODE.

ChrisRackauckas:
Yeah, that would make sense. It would then match well with our collocation tools:

https://diffeqflux.sciml.ai/dev/examples/collocation/
https://diffeqflux.sciml.ai/dev/Collocation/

We should also do a non-smoothed collocation method via DataInterpolations, and demonstrate that on this HamiltonianNN

ChrisRackauckas:
Also, just a note, some of the models from the HNN paper are impossible to fit directly from the solve on the time-series. Specifically the double pendulum, since it's chaotic so the adjoints diverge exponentially fast.

avik-pal:
ReverseDiff seems to bug out on GPU with `CuArray only supports bits types`.

```
hnn = HamiltonianNN(
    Chain(Dense(2, 64, relu), Dense(64, 1)) |> gpu
)

p = hnn.p;

hnn(rand(2, 1) |> gpu)

ReverseDiff.gradient(p -> sum(hnn(rand(2, 1) |> gpu, p)), p)
```

ChrisRackauckas:
>ReverseDiff seems to bug out on GPU with CuArray only supports bits types.

Yes it does. I think we'll need to just skip that for now, open an issue, and grab what we can get. Keno's AD should fix this.

Vaibhavdixit02:
Looks great! 🎉

avik-pal:
> We should also do a non-smoothed collocation method via DataInterpolations, and demonstrate that on this HamiltonianNN

@ChrisRackauckas could you elaborate a bit on this?

ChrisRackauckas:
Essentially one way to perform this fitting is to fit spliens like a cublic spline from DataInterpolations, and then take the derivative of the spline to give you estimated derivatives, and then perform the fit directly against those derivative approximations. It would be nice to have this version under the same interface as the smoothed one, which is essentially the same idea except using regression splines under some smoothing kernel to get derivatives of very noisy and dense data.

PR #371, opened 2020-07-21
new dispatch

d-netto:


src/neural_de.jl
```diff
@@ -443,6 +443,15 @@ struct NeuralODEMM{M,M2,P,RE,T,MM,A,K} <: NeuralDELayer
             typeof(tspan),typeof(mass_matrix),typeof(args),typeof(kwargs)}(
             model,constraints_model,p,re,tspan,mass_matrix,args,kwargs)
     end
+
+    function NeuralODEMM(model::TensorLayer,constraints_model,tspan,mass_matrix,args...;
```
ChrisRackauckas, line 447: shouldn't the TensorLayer be used in a FastChain?

d-netto: That's true. The model that should be a TensorLayer.

d-netto: I don't think so. I think it should be the model itself.

PR #372, opened 2020-07-24
Regularized ffjord [WIP]

d-netto:


d-netto:
Regularization seems t be making it worse. Probably an issue with the signs in the ODE.

src/ffjord.jl
```diff
@@ -1,11 +1,35 @@
 abstract type CNFLayer <: Function end
 Flux.trainable(m::CNFLayer) = (m.p,)
 
+struct DeterministicCNFLayer{M,P,RE,Distribution,T,A,K} <: CNFLayer
```
ChrisRackauckas, line 4: the other layers don't have this Layer in the name. For consistency it should be dropped.

src/ffjord.jl
```diff
@@ -61,25 +82,35 @@ function jacobian_fn(f, x::AbstractVector)
...
         trace_jac = (eJ.*e)[1]
     else
         J = jacobian_fn(m, z)
         trace_jac = length(z) == 1 ? sum(J) : tr(J)
     end
-    du[1:end-1] = m(z)
-    du[end] = -trace_jac
+    du[1:end-3] = m(z)
+    du[end-2] = -trace_jac
+    du[end-1] = sum(abs2, m(z))
```
ChrisRackauckas, line 107: so FFJORD would always regularize? Seems like that should be an option.

src/ffjord.jl
d-netto: Ok.

src/ffjord.jl
d-netto: Yeah, in the current implementation it's always regularized. I guess we can put a switch to enable regularization here.

d-netto:
TODO: Write documentation for DeterministicCNF

PR #380, opened 2020-08-02
[WIP] Extend collocation methods to use DataInterpolations.jl

avik-pal:
Follows from discussion on Non-Smooth Collocation Methods in #370


src/collocation.jl
```diff
@@ -140,3 +140,27 @@ function collocate_data(data,tpoints,kernel=TriangularKernel())
   estimated_solution = reduce(hcat,transpose.(last.(x)))
   estimated_derivative,estimated_solution
 end
+
+# FIXME: Dispatch using AbstractInterpolation not working :(
```
ChrisRackauckas, line 144: what's the issue?

avik-pal: I just realized that I was passing it as `CubicSpline` instead of `CubicSpline()` so it is expected to fail. However, I am wondering how to make it all fit into a common API with the Smoothened Kernel ones? Should we just have the additional `tpoints_sample` argument differentiating the functions or maybe create a wrapper around the DataInterpolations types?

ChrisRackauckas: I think a wrapper over DataInterpolations types might be necessary. But also, directly defining the derivatives of its interoplations might be helpful too, since for polynomials we can just directly know the derivatives of the interpolations.

avik-pal: Do you mean not using ForwardDiff for the derivatives and have those functions manually defined?

ChrisRackauckas: Yes, that might make all of the differentiation much easier.

ChrisRackauckas:
Okay, this is interesting. It seems that the DataInterpolations PR broke master and it's the same error error. The test in question wants to use the derivative w.r.t. the data of the interpolation. 

https://github.com/PumasAI/DataInterpolations.jl/blob/master/src/DataInterpolations.jl#L24-L27

That `DoesNotExist()` I think should be interpreted as the derivative of the output w.r.t. the parameters. The thing acts as an array, https://github.com/PumasAI/DataInterpolations.jl/blob/master/src/DataInterpolations.jl#L9-L10 so I think it just needs to send back an array of mostly zeros and the local stencil coefficients in the right place.

PR #387, opened 2020-08-10
Convolution MNIST example

jessebett:
@ChrisRackauckas This is a mostly working example modified from the existing demo.


docs/src/examples/conv_mnist_neural_ode.md
```diff
@@ -0,0 +1,427 @@
...
+end
+
+# Main
+const bs = 128
+const train_split = 0.9
+train_dataloader, test_dataloader = loadmnist(bs, train_split);
+
+down = Chain(
+             Conv((3,3),1=>64,relu,stride=1), GroupNorm(64,64),
+             Conv((4,4),64=>64,relu,stride=2,pad=1), GroupNorm(64,64),
```
ChrisRackauckas, line 45: interesting. All of these Convs fit on tensor cores. I wonder if they are used.

DhairyaLGandhi: Tensor cores would be used with the incoming changes into the CUDNN wrappers, which haven't been merged yet

ChrisRackauckas:
@DhairyaLGandhi do you know why the data loader is trying to load like 9gb onto the GPU here?

DhairyaLGandhi:
It might be trying to move the whole dataset over to the GPU at once.

ChrisRackauckas:
```julia
using DiffEqFlux, OrdinaryDiffEq, Flux, NNlib,  Printf
using Flux: logitcrossentropy
using Flux.Data: DataLoader
using MLDatasets
using MLDataUtils:  LabelEnc, convertlabel, stratifiedobs
using CUDA
CUDA.allowscalar(false)

function loadmnist(batchsize = bs, train_split = 0.9)
    # Use MLDataUtils LabelEnc for natural onehot conversion
    onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw,
                                      LabelEnc.NativeLabels(collect(0:9)))
    # Load MNIST
    imgs, labels_raw = MNIST.traindata();
    # Process images into (H,W,C,BS) batches
    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
    y_data = onehot(labels_raw)
    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
                                                         p = train_split)
    return (
        # Use Flux's DataLoader to automatically minibatch and shuffle the data
        DataLoader(collect.([x_train, y_train]); batchsize = batchsize,
                   shuffle = true),
        # Don't shuffle the test data
        DataLoader(collect.([x_test, y_test]); batchsize = batchsize,
                   shuffle = false)
    )
end

# Main
const bs = 128
const train_split = 0.9
train_dataloader, test_dataloader = loadmnist(bs, train_split);

down = Chain(
             Conv((3,3),1=>64,relu,stride=1), GroupNorm(64,64),
             Conv((4,4),64=>64,relu,stride=2,pad=1), GroupNorm(64,64),
             Conv((4,4),64=>64,stride=2,pad=1),
            )|>gpu;
dudt = Chain(
           Conv((3,3),64=>64,relu,stride=1,pad=1),
           Conv((3,3),64=>64,relu,stride=1,pad=1)
          ) |>gpu;

fc = Chain(GroupNorm(64,64),
           x->relu.(x),
           MeanPool((6,6)),
           x -> reshape(x,(64,)),
           Dense(64,10)
          )|>gpu;

nn_ode = NeuralODE(dudt, (0.f0, 1.f0), Tsit5(),
                   save_everystep = false,
                   reltol = 1e-3, abstol = 1e-3,
                   save_start = false) |> gpu


function DiffEqArray_to_Array(x)
    xarr = gpu(x)
    return xarr[:,:,:,1,:]
end

# Build our over-all model topology
model = Chain(
              down,             #(28,28,1,BS) -> (6,6,64,BS)
              nn_ode,           #(6,6,64,BS) -> (6,6,64,BS)
              DiffEqArray_to_Array,
              fc                #(6,6,64,BS) -> (10, BS)
             )

# To understand the intermediate NN-ODE layer, we can examine it's dimensionality
img, lab = train_dataloader.data[1][:, :, :, 1:1], train_dataloader.data[2][:, 1:1]

x_d = down(gpu(img))

# We can see that we can compute the forward pass through the NN topology
# featuring an NNODE layer.
x_m = model(gpu(img))

classify(x) = argmax.(eachcol(x))

function accuracy(model, data; n_batches = 100)
    total_correct = 0
    total = 0
    for (i, (x, y)) in enumerate(data)
        # Only evaluate accuracy for n_batches
        i > n_batches && break
        target_class = classify(cpu(y))
        predicted_class = classify(cpu(model(x)))
        total_correct += sum(target_class .== predicted_class)
        total += length(target_class)
    end
    return total_correct / total
end

# burn in accuracy
accuracy(model, train_dataloader)

loss(x, y) = logitcrossentropy(model(gpu(x)), gpu(y))

# burn in loss
loss(img, lab)

opt = ADAM(0.05)
iter = 0

cb() = begin
    global iter += 1
    # Monitor that the weights do infact update
    # Every 10 training iterations show accuracy
    if iter % 10 == 1
        train_accuracy = accuracy(model, train_dataloader) * 100
        test_accuracy = accuracy(model, test_dataloader;
                                 n_batches = length(test_dataloader)) * 100
        @printf("Iter: %3d || Train Accuracy: %2.3f || Test Accuracy: %2.3f\n",
                iter, train_accuracy, test_accuracy)
    end
end

# Train the NN-ODE and monitor the loss and weights.
Flux.train!(loss, params(down, nn_ode.p, fc), train_dataloader, opt, cb = cb)
```

is what I'm trying, but why would the other example be fine? Maybe @avik-pal knows this stuff.

avik-pal:
Its not clear why DataLoader would take that much GPU memory. We keep the data on CPU and move it to GPU on every iteration `loss(x, y) = logitcrossentropy(model(gpu(x)), gpu(y))`

ChrisRackauckas:
My version is okay on GPU memory. However, it seems to fail in other ways. But @jessebett 's example is a lot like the example we have in the tutorials, and that is fine on GPU memory, so I'm not sure what's different.

avik-pal:
```julia
function loadmnist(batchsize = bs, train_split = 0.9)
    # Use MLDataUtils LabelEnc for natural onehot conversion
    onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw,
                                      LabelEnc.NativeLabels(collect(0:9)))
    # Load MNIST
    imgs, labels_raw = MNIST.traindata();
    # Process images into (H,W,C,BS) batches
    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
    y_data = onehot(labels_raw)
    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
                                                         p = train_split)
    return (
        # Use Flux's DataLoader to automatically minibatch and shuffle the data
        DataLoader(collect.([x_train, y_train])...; batchsize = batchsize,
                   shuffle = true),
        # Don't shuffle the test data
        DataLoader(collect.([x_test, y_test])...; batchsize = batchsize,
                   shuffle = false)
    )
end

function accuracy(model, data; n_batches = 100)
    total_correct = 0
    total = 0
    for (i, (x, y)) in enumerate(data)
        # Only evaluate accuracy for n_batches
        i > n_batches && break
        target_class = classify(cpu(y))
        predicted_class = classify(cpu(model(gpu(x))))
        total_correct += sum(target_class .== predicted_class)
        total += length(target_class)
    end
    return total_correct / total
end
```

This should fix the code.

ChrisRackauckas:
@avik-pal did that work for you? I am using

```julia
using DiffEqFlux, OrdinaryDiffEq, Flux, NNlib,  Printf
using Flux: logitcrossentropy
using Flux.Data: DataLoader
using MLDatasets
using MLDataUtils:  LabelEnc, convertlabel, stratifiedobs
using CUDA, DiffEqSensitivity
CUDA.allowscalar(false)

function loadmnist(batchsize = bs, train_split = 0.9)
    # Use MLDataUtils LabelEnc for natural onehot conversion
    onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw,
                                      LabelEnc.NativeLabels(collect(0:9)))
    # Load MNIST
    imgs, labels_raw = MNIST.traindata();
    # Process images into (H,W,C,BS) batches
    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
    y_data = onehot(labels_raw)
    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
                                                         p = train_split)
    return (
        # Use Flux's DataLoader to automatically minibatch and shuffle the data
        DataLoader(collect.([x_train, y_train])...; batchsize = batchsize,
                   shuffle = true),
        # Don't shuffle the test data
        DataLoader(collect.([x_test, y_test])...; batchsize = batchsize,
                   shuffle = false)
    )
end

# Main
const bs = 64
const train_split = 0.9
train_dataloader, test_dataloader = loadmnist(bs, train_split);

down = Chain(
             Conv((3,3),1=>64,relu,stride=1), GroupNorm(64,64),
             Conv((4,4),64=>64,relu,stride=2,pad=1), GroupNorm(64,64),
             Conv((4,4),64=>64,stride=2,pad=1),
            )|>gpu;
dudt = Chain(
           Conv((3,3),64=>64,relu,stride=1,pad=1),
           Conv((3,3),64=>64,relu,stride=1,pad=1)
          ) |>gpu;

fc = Chain(GroupNorm(64,64),
           x->relu.(x),
           MeanPool((6,6)),
           x -> reshape(x,(64,)),
           Dense(64,10)
          )|>gpu;

nn_ode = NeuralODE(dudt, (0.f0, 1.f0), Tsit5(),
                   save_everystep = false,
                   reltol = 1e-3, abstol = 1e-3,
                   sensealg = BacksolveAdjoint(),
                   save_start = false) |> gpu


function DiffEqArray_to_Array(x)
    xarr = gpu(x)
    return xarr[:,:,:,1,:]
end

# Build our over-all model topology
model = Chain(
              down,             #(28,28,1,BS) -> (6,6,64,BS)
              nn_ode,           #(6,6,64,BS) -> (6,6,64,BS)
              DiffEqArray_to_Array,
              fc                #(6,6,64,BS) -> (10, BS)
             )

# To understand the intermediate NN-ODE layer, we can examine it's dimensionality
img, lab = train_dataloader.data[1][:, :, :, 1:1], train_dataloader.data[2][:, 1:1]

x_d = down(gpu(img))

# We can see that we can compute the forward pass through the NN topology
# featuring an NNODE layer.
x_m = model(gpu(img))

classify(x) = argmax.(eachcol(x))

function accuracy(model, data; n_batches = 100)
    total_correct = 0
    total = 0
    for (i, (x, y)) in enumerate(data)
        # Only evaluate accuracy for n_batches
        i > n_batches && break
        target_class = classify(cpu(y))
        predicted_class = classify(cpu(model(gpu(x))))
        total_correct += sum(target_class .== predicted_class)
        total += length(target_class)
    end
    return total_correct / total
end

# burn in accuracy
accuracy(model, train_dataloader)

loss(x, y) = logitcrossentropy(model(gpu(x)), gpu(y))

# burn in loss
loss(img, lab)

opt = ADAM(0.05)
iter = 0

cb() = begin
    global iter += 1
    # Monitor that the weights do infact update
    # Every 10 training iterations show accuracy
    if iter % 10 == 1
        train_accuracy = accuracy(model, train_dataloader) * 100
        test_accuracy = accuracy(model, test_dataloader;
                                 n_batches = length(test_dataloader)) * 100
        @printf("Iter: %3d || Train Accuracy: %2.3f || Test Accuracy: %2.3f\n",
                iter, train_accuracy, test_accuracy)
    end
end

# Train the NN-ODE and monitor the loss and weights.
Flux.train!(loss, params(down, nn_ode.p, fc), train_dataloader, opt, cb = cb)
```

and it's still taking quite a bit of memory for me, and I'm hitting a deprecation warning from Flux:

```julia
┌ Warning: `DataLoader(x...; kws...)` is deprecated, use `DataLoader(x; kws...)` instead.
│   caller = ip:0x0
└ @ Core :-1
```

DhairyaLGandhi:
You want to use a tuple to construct the Data loader instead of the array, and avoid the splatting there

ChrisRackauckas:
I'm getting OOM again when I do

```julia
function loadmnist(batchsize = bs, train_split = 0.9)
    # Use MLDataUtils LabelEnc for natural onehot conversion
    onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw,
                                      LabelEnc.NativeLabels(collect(0:9)))
    # Load MNIST
    imgs, labels_raw = MNIST.traindata();
    # Process images into (H,W,C,BS) batches
    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
    y_data = onehot(labels_raw)
    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
                                                         p = train_split)
    return (
        # Use Flux's DataLoader to automatically minibatch and shuffle the data
        DataLoader(collect.((x_train, y_train)); batchsize = batchsize,
                   shuffle = true),
        # Don't shuffle the test data
        DataLoader(collect.((x_test, y_test)); batchsize = batchsize,
                   shuffle = false)
    )
end

```

ChrisRackauckas:
```julia
using DiffEqFlux, OrdinaryDiffEq, Flux, NNlib,  Printf
using Flux: logitcrossentropy
using Flux.Data: DataLoader
using MLDatasets
using MLDataUtils:  LabelEnc, convertlabel, stratifiedobs
using CUDA
CUDA.allowscalar(false)

function loadmnist(batchsize = bs, train_split = 0.9)
    # Use MLDataUtils LabelEnc for natural onehot conversion
    onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw,
                                      LabelEnc.NativeLabels(collect(0:9)))
    # Load MNIST
    imgs, labels_raw = MNIST.traindata();
    # Process images into (H,W,C,BS) batches
    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
    y_data = onehot(labels_raw)
    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
                                                         p = train_split)
    return (
        # Use Flux's DataLoader to automatically minibatch and shuffle the data
        DataLoader(gpu.(collect.((x_train, y_train))); batchsize = batchsize,
                   shuffle = true),
        # Don't shuffle the test data
        DataLoader(gpu.(collect.((x_test, y_test))); batchsize = batchsize,
                   shuffle = false)
    )
end

# Main
const bs = 8
const train_split = 0.9
train_dataloader, test_dataloader = loadmnist(bs, train_split);

down = Chain(
             Conv((3,3),1=>64,relu,stride=1), GroupNorm(64,64),
             Conv((4,4),64=>64,relu,stride=2,pad=1), GroupNorm(64,64),
             Conv((4,4),64=>64,stride=2,pad=1),
            )|>gpu;
dudt = Chain(
           Conv((3,3),64=>64,relu,stride=1,pad=1),
           Conv((3,3),64=>64,relu,stride=1,pad=1)
          ) |>gpu;

fc = Chain(GroupNorm(64,64),
           x->relu.(x),
           MeanPool((6,6)),
           x -> reshape(x,(64,)),
           Dense(64,10)
          )|>gpu;

nn_ode = NeuralODE(dudt, (0.f0, 1.f0), Tsit5(),
                   save_everystep = false,
                   reltol = 1e-3, abstol = 1e-3,
                   save_start = false) |> gpu


function DiffEqArray_to_Array(x)
    xarr = gpu(x)
    return xarr[:,:,:,1,:]
end

# Build our over-all model topology
model = Chain(
              down,             #(28,28,1,BS) -> (6,6,64,BS)
              nn_ode,           #(6,6,64,BS) -> (6,6,64,BS)
              DiffEqArray_to_Array,
              fc                #(6,6,64,BS) -> (10, BS)
             )

# To understand the intermediate NN-ODE layer, we can examine it's dimensionality
img, lab = train_dataloader.data[1][:, :, :, 1:1], train_dataloader.data[2][:, 1:1]

x_d = down(img)

# We can see that we can compute the forward pass through the NN topology
# featuring an NNODE layer.
x_m = model(img)

classify(x) = argmax.(eachcol(x))

function accuracy(model, data; n_batches = 100)
    total_correct = 0
    total = 0
    for (i, (x, y)) in enumerate(data)
        # Only evaluate accuracy for n_batches
        i > n_batches && break
        target_class = classify(cpu(y))
        predicted_class = classify(cpu(model(x)))
        total_correct += sum(target_class .== predicted_class)
        total += length(target_class)
    end
    return total_correct / total
end

# burn in accuracy
accuracy(model, train_dataloader)

loss(x, y) = logitcrossentropy(model(x), y)

# burn in loss
loss(img, lab)

opt = ADAM(0.05)
iter = 0

function cb()
    global iter += 1
    # Monitor that the weights do infact update
    # Every 10 training iterations show accuracy
    if iter % 10 == 1
        train_accuracy = accuracy(model, train_dataloader) * 100
        test_accuracy = accuracy(model, test_dataloader;
                                 n_batches = length(test_dataloader)) * 100
        @printf("Iter: %3d || Train Accuracy: %2.3f || Test Accuracy: %2.3f\n",
                iter, train_accuracy, test_accuracy)
    end
end

# Train the NN-ODE and monitor the loss and weights.
Flux.train!(loss, params(down, nn_ode.p, fc), train_dataloader, opt, cb = cb)
```

Goes OOM.

avik-pal:
The callback to accuracy seems to allocate around 5GB per call which is never freed. Removing the callback I get `Instability Detected; Aborting`

ChrisRackauckas:
The ADAM learning rate was too high so it was diverging. This is much better:

```julia
using DiffEqFlux, OrdinaryDiffEq, Flux, NNlib,  Printf
using Flux: logitcrossentropy
using Flux.Data: DataLoader
using MLDatasets
using MLDataUtils:  LabelEnc, convertlabel, stratifiedobs
using CUDA
CUDA.allowscalar(false)

function loadmnist(batchsize = bs, train_split = 0.9)
    # Use MLDataUtils LabelEnc for natural onehot conversion
    onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw,
                                      LabelEnc.NativeLabels(collect(0:9)))
    # Load MNIST
    imgs, labels_raw = MNIST.traindata();
    # Process images into (H,W,C,BS) batches
    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
    y_data = onehot(labels_raw)
    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
                                                         p = train_split)
    return (
        # Use Flux's DataLoader to automatically minibatch and shuffle the data
        DataLoader(gpu.(collect.((x_train, y_train))); batchsize = batchsize,
                   shuffle = true),
        # Don't shuffle the test data
        DataLoader(gpu.(collect.((x_test, y_test))); batchsize = batchsize,
                   shuffle = false)
    )
end

# Main
const bs = 128
const train_split = 0.9
train_dataloader, test_dataloader = loadmnist(bs, train_split);

down = Chain(
             Conv((3,3),1=>64,relu,stride=1), GroupNorm(64,64),
             Conv((4,4),64=>64,relu,stride=2,pad=1), GroupNorm(64,64),
             Conv((4,4),64=>64,stride=2,pad=1),
            )|>gpu;
dudt = Chain(
           Conv((3,3),64=>64,relu,stride=1,pad=1),
           Conv((3,3),64=>64,relu,stride=1,pad=1)
          ) |>gpu;

fc = Chain(GroupNorm(64,64),
           x->relu.(x),
           MeanPool((6,6)),
           x -> reshape(x,(64,)),
           Dense(64,10)
          )|>gpu;

nn_ode = NeuralODE(dudt, (0.f0, 1.f0), Tsit5(),
                   save_everystep = false,
                   reltol = 1e-3, abstol = 1e-3,
                   save_start = false) |> gpu


function DiffEqArray_to_Array(x)
    xarr = gpu(x)
    return xarr[:,:,:,1,:]
end

# Build our over-all model topology
model = Chain(
              down,             #(28,28,1,BS) -> (6,6,64,BS)
              nn_ode,           #(6,6,64,BS) -> (6,6,64,BS)
              DiffEqArray_to_Array,
              fc                #(6,6,64,BS) -> (10, BS)
             )

# To understand the intermediate NN-ODE layer, we can examine it's dimensionality
img, lab = train_dataloader.data[1][:, :, :, 1:1], train_dataloader.data[2][:, 1:1]

x_d = down(img)

# We can see that we can compute the forward pass through the NN topology
# featuring an NNODE layer.
x_m = model(img)

classify(x) = argmax.(eachcol(x))

function accuracy(model, data; n_batches = 100)
    total_correct = 0
    total = 0
    for (i, (x, y)) in enumerate(data)
        # Only evaluate accuracy for n_batches
        i > n_batches && break
        target_class = classify(cpu(y))
        predicted_class = classify(cpu(model(x)))
        total_correct += sum(target_class .== predicted_class)
        total += length(target_class)
    end
    return total_correct / total
end

# burn in accuracy
accuracy(model, train_dataloader)

loss(x, y) = logitcrossentropy(model(x), y)

# burn in loss
loss(img, lab)

opt = ADAM(0.001)
iter = 0

function cb()
    global iter += 1
    # Monitor that the weights do infact update
    # Every 10 training iterations show accuracy
    print(iter)
    print(" ")
    println(loss(img, lab))
end

# Train the NN-ODE and monitor the loss and weights.
Flux.train!(loss, params(down, nn_ode.p, fc), train_dataloader, opt, cb=cb)
```

That seems to work, but I can't seem to print losses on any test data without getting memory issues.

RobWalt:
I'm pretty interested in this specific example. What's the current status?

The latest version didn't work for me until I changed some things:

- batch size of 128 leads to the OOM Error on my machine, so I used 8 as suggested in an earlier version
- In the definition of `fc` I had to add a colon in the reshape line, so it reads `x -> reshape(x, (64,:))`
- In `DiffEqArray_to_Array` we have solution dimensions [W,H,C,B,T] (right?). So we need to access via `xarr[:,:,:,:,1]` opposed to `xarr[:,:,:,1,:]`

After all of this, the example ran. It seems that there is no convergence however. Additionally it runs very slowly. I already experimented a bit with the ADAM learning rate and higher rates seem to slow the process down quite a bit, while not improving the result at all.

ChrisRackauckas:
I'm pretty sure something weird with Flux is happening, specifically in the DataLoader memory usage, so I think someone like @DhairyaLGandhi would need to take a look at it.

RobWalt:
Yeah, I think you're right. Yesterday I coded a version without DataLoader. Some parts are a bit wonky, maybe. But here it is and it works just fine without any problems up to this point.

```julia
using Flux
using OrdinaryDiffEq
using DiffEqFlux
using CUDA
CUDA.allowscalar(false)

# function for making an input data batch
function make_input_batch(start_index, finish_index)
  cat( (map(x -> Float32.(x), MNIST_img[start_index:finish_index]))... ,dims=4 )
end

# onehot encoding for a batch of data
function onehot_batch(to_onehot_data)
  cat( [Flux.onehot(x, 0:9) for x in to_onehot_data]..., dims=2 )
end

# load the data
MNIST_img = Flux.Data.MNIST.images()
MNIST_lab = Flux.Data.MNIST.labels()

# prepare train/test data
batch_size = 128
train_end_point = 50000
num_batches = train_end_point ÷ batch_size
train_inp = gpu([make_input_batch(1+batch_size*i, batch_size*(i+1)) for i in 1:num_batches])
train_lab = gpu([onehot_batch(MNIST_lab[1+batch_size*i:batch_size*(i+1)]) for i in 1:num_batches])
test_in = gpu(make_input_batch(train_end_point, train_end_point+5000))
test_lab = gpu(onehot_batch(MNIST_lab[train_end_point:train_end_point+5000]))
train_data = gpu(zip(train_inp, train_lab))

# model definition
down = Chain(
  Conv((3,3), 1 => 32, relu, stride=2, pad=1),
  GroupNorm(32,32),
  Conv((3,3), 32 => 32, relu, stride=2, pad=1),
  GroupNorm(32,32),
  Conv((3,3), 32 => 32, stride=2, pad=1)
  ) |>gpu

dudt = Chain(
        Conv((3,3), 32 => 32, stride=1, pad=1),
        Conv((3,3), 32 => 32, stride=1, pad=1)
        ) |>gpu

conv_ode = Chain(
  NeuralODE(
    dudt,
    (0.f0, 1.f0),
    Tsit5(),
    save_everystep=false,
    rel_tol=1e-3,
    abs_tol=1e-3,
    save_start=false),
  x -> gpu(x),
  x -> x[:,:,:,:,1]
          ) |>gpu

res = Chain(
  flatten,
  Dense(4*4*32, 10),
  softmax
  ) |>gpu

conv_net = Chain(down, conv_ode, res) |> gpu

# test the model with one input batch
x_in = train_inp[1]

x_d = down(x_in)
x_ode = conv_ode(x_d)
x_out = res(x_ode)

x_full_out = conv_net(x_in)

# define loss function
function mnist_loss(img, true_preds)
  pred_vecs = conv_net(img)
  Flux.Losses.crossentropy(pred_vecs, true_preds)
end

# define optimiser
opt = Flux.Optimise.ADAM(0.0005)

# get model parameters
parameters = Flux.params(conv_net)

# define callback for feedback while training
iter = 0
function eval_mnist(test_x, test_y)
  global iter += 1
  println("Progress: $iter/$num_batches")
  if iter % 10 == 0
    @time(@show(mnist_loss(test_x, test_y)))
  end
end

evalcb() = eval_mnist(test_in, test_lab)

# train the net
Flux.Optimise.train!(mnist_loss, parameters, train_data, opt, cb=evalcb)
```

By the way quick question: What's the advantage of using DataLoader? Is it just a function for convenience or is there some kind of optimisation happening in the background when using it?

ToucheSir:
@RobWalt that fails for me with a stackoverflow in `make_input_batch`. I adapted @ChrisRackauckas's latest version to use `DataLoader` and it seems to not OOM:
```julia
## imports
using Flux
using Flux: logitcrossentropy
using Flux.Data: DataLoader
using Printf
using MLDatasets
using OrdinaryDiffEq
using DiffEqFlux
using CUDA
CUDA.allowscalar(false)

## load the data
function loadmnist(batchsize)
    train_x = Flux.unsqueeze(MNIST.traintensor(Float32), 3)
    test_x = Flux.unsqueeze(MNIST.testtensor(Float32), 3)
    train_y = Float32.(Flux.onehotbatch(MNIST.trainlabels(), 0:9))
    test_y = Float32.(Flux.onehotbatch(MNIST.testlabels(), 0:9))

    train_loader = DataLoader((train_x, train_y), batchsize=batchsize, shuffle=true)
    test_loader = DataLoader((test_x, test_y), batchsize=batchsize, shuffle=false)
    return train_loader, test_loader
end

const bs = 128
train_dataloader, test_dataloader = loadmnist(bs);

## model definition
down = Chain(
             Conv((3,3),1=>64,relu,stride=1), GroupNorm(64,64),
             Conv((4,4),64=>64,relu,stride=2,pad=1), GroupNorm(64,64),
             Conv((4,4),64=>64,stride=2,pad=1),
            )|>gpu;
dudt = Chain(
           Conv((3,3),64=>64,relu,stride=1,pad=1),
           Conv((3,3),64=>64,relu,stride=1,pad=1)
          ) |>gpu;

fc = Chain(GroupNorm(64,64),
           x->relu.(x),
           MeanPool((6,6)),
           x -> reshape(x,(64,)),
           Dense(64,10)
          )|>gpu;

nn_ode = NeuralODE(dudt, (0.f0, 1.f0), Tsit5(),
                   save_everystep = false,
                   reltol = 1e-3, abstol = 1e-3,
                   save_start = false) |> gpu


function DiffEqArray_to_Array(x)
    xarr = gpu(x)
    return xarr[:,:,:,1,:]
end

# Build our over-all model topology
model = Chain(
              down,             #(28,28,1,BS) -> (6,6,64,BS)
              nn_ode,           #(6,6,64,BS) -> (6,6,64,BS)
              DiffEqArray_to_Array,
              fc                #(6,6,64,BS) -> (10, BS)
             )

## To understand the intermediate NN-ODE layer, we can examine it's dimensionality
img, lab = train_dataloader.data[1][:, :, :, 1:1], train_dataloader.data[2][:, 1:1]

x_d = down(gpu(img))

# We can see that we can compute the forward pass through the NN topology
# featuring an NNODE layer.
x_m = model(gpu(img))

## loss
loss(x, y) = logitcrossentropy(model(x), y)
loss(gpu(img), gpu(lab))

## accuracy
classify(x) = argmax.(eachcol(x))

function accuracy(model, data; n_batches = 100)
    total_correct = 0
    total = 0
    for (i, (x, y)) in enumerate(data)
        # Only evaluate accuracy for n_batches
        i > n_batches && break
        target_class = classify(cpu(y))
        predicted_class = classify(cpu(model(x)))
        total_correct += sum(target_class .== predicted_class)
        total += length(target_class)
    end
    return total_correct / total
end

accuracy(model, CuIterator(train_dataloader))
CUDA.reclaim()

## training setup
opt = ADAM(0.001)
iter = 0

cb() = begin
    global iter += 1
    # Monitor that the weights do infact update
    # Every 10 training iterations show accuracy
    if iter % 10 == 1
        train_accuracy = accuracy(model, CuIterator(train_dataloader)) * 100
        test_accuracy = accuracy(model, CuIterator(test_dataloader);
                                 n_batches = length(test_dataloader)) * 100
        @printf("Iter: %3d || Train Accuracy: %2.3f || Test Accuracy: %2.3f\n",
                iter, train_accuracy, test_accuracy)
        # CUDA.reclaim()
        CUDA.memory_status()
    end
end

## Train the NN-ODE and monitor the loss and weights.
Flux.train!(loss, params(down, nn_ode.p, fc), CuIterator(train_dataloader), opt, cb = cb)
```
I also tried with pre-loading everything onto GPU with `gpu.(train/test_dataloader)` and the performance is comparable. `CUDA.reclaim()` is not necessary to avoid OOMs, but it might clear up some cruft that accumulates during the callback. Based on https://github.com/JuliaGPU/CUDA.jl/issues/149, Flux may be a culprit. VRAM usage jumps from ~900MB (mostly model params) to ~8GB/100% in the first 10 batches. Based on how much the compute usage varies (even without the callback) it appears like the GC is having some difficulty keeping up.

darsnack:
At @ChrisRackauckas request on the last ML call, I wanted to try and address some of the OOM errors in this thread. Thankfully, @ToucheSir beat me to mentioning `CuIterator` which should help with GC on the GPU side. Another addition is [DataLoaders.jl](https://github.com/lorenzoh/DataLoaders.jl) from the ML community team which I hope will supersede Flux.DataLoader in the long term. A key aspect of its design is that it follows MLDataPattern.jl's interfaces closely, so it is plug-n-play with most of your JuliaML functions, and it adheres to the "lazy wrapper" behavior of most of MLDataPattern.jl. By default, it also makes use of the additional CPU threads to load data in the background. Here's the code snippet:
```julia
using DiffEqFlux, OrdinaryDiffEq, Flux, NNlib,  Printf
using Flux: logitcrossentropy
using DataLoaders
using MLDatasets
using MLDataPattern
using LearnBase
using MLDataUtils:  LabelEnc, convertlabel
using CUDA
CUDA.allowscalar(false)

# Use MLDataUtils LabelEnc for natural onehot conversion
array_wrap(x::AbstractArray) = x
array_wrap(x::Number) = [x]
onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw, collect(0:9))

# implement a MLDataPattern compatible interface
# long-term we will update MLDatasets to do this automatically
struct MNISTDataset{T, S}
    imgs::T
    labels::S
end
MNISTDataset() = MNISTDataset(MNIST.traindata(Float32)...)
LearnBase.nobs(d::MNISTDataset) = size(d.imgs, 3)
function LearnBase.getobs(d::MNISTDataset, idx)
    imgs, labels = d.imgs[:, :, idx], d.labels[idx]

    return reshape(imgs, size(imgs, 1), size(imgs, 2), 1), vec(onehot(array_wrap(labels)))
end

# loadmnist is now returning lazy loaders
# the data is only read when the getobs call occurs
# this could be extended to only loading data from disk
function loadmnist(batchsize = bs, train_split = 0.9)
    dataset = MNISTDataset() # reference our nobs and getobs
    traindata, valdata = MLDataPattern.splitobs(dataset; at = train_split) # split training data

    return (
        # Use DataLoaders.DataLoader instead
        DataLoader(shuffleobs(traindata), batchsize),
        # Don't shuffle the test data
        DataLoader(valdata, batchsize)
    )
end

# Main
const bs = 128
const train_split = 0.9
train_dataloader, test_dataloader = loadmnist(bs, train_split);

down = Chain(Conv((3,3),1=>64,relu,stride=1), GroupNorm(64,64),
             Conv((4,4),64=>64,relu,stride=2,pad=1), GroupNorm(64,64),
             Conv((4,4),64=>64,stride=2,pad=1)) |> gpu;
dudt = Chain(Conv((3,3),64=>64,relu,stride=1,pad=1),
             Conv((3,3),64=>64,relu,stride=1,pad=1)) |> gpu;

fc = Chain(GroupNorm(64,64),
           x->relu.(x),
           MeanPool((6,6)),
           x -> reshape(x,(64,)),
           Dense(64,10)) |> gpu;

nn_ode = NeuralODE(dudt, (0.f0, 1.f0), Tsit5(),
                   save_everystep = false,
                   reltol = 1e-3, abstol = 1e-3,
                   save_start = false) |> gpu;


function DiffEqArray_to_Array(x)
    xarr = gpu(x)
    return xarr[:,:,:,1,:]
end

# Build our over-all model topology
model = Chain(down,             #(28,28,1,BS) -> (6,6,64,BS)
              nn_ode,           #(6,6,64,BS) -> (6,6,64,BS)
              DiffEqArray_to_Array,
              fc)               #(6,6,64,BS) -> (10, BS)

# To understand the intermediate NN-ODE layer, we can examine it's dimensionality
img, lab = first(train_dataloader) .|> gpu;

x_d = down(img)

# We can see that we can compute the forward pass through the NN topology
# featuring an NNODE layer.
x_m = model(img)

classify(x) = argmax.(eachcol(x))

function accuracy(model, data; n_batches = 100)
    total_correct = 0
    total = 0
    for (i, (x, y)) in enumerate(data)
        # Only evaluate accuracy for n_batches
        i > n_batches && break
        target_class = classify(cpu(y))
        predicted_class = classify(cpu(model(x)))
        total_correct += sum(target_class .== predicted_class)
        total += length(target_class)
    end
    return total_correct / total
end

loss(x, y) = logitcrossentropy(model(x), y)

# burn in loss
# do this before accuracy
# iterating train_dataloader will clear the buffers of img, lab
loss(img, lab)

# burn in accuracy
accuracy(model, CuIterator(train_dataloader))

opt = ADAM(0.001)
iter = 0

function cb()
    global iter += 1
    # Monitor that the weights do infact update
    # Every 10 training iterations show accuracy
    if iter % 100 == 1
        # train_accuracy = accuracy(model, CuIterator(train_dataloader)) * 100
        test_accuracy = accuracy(model, CuIterator(test_dataloader);
                                 n_batches = nobs(test_dataloader.data)) * 100
        @printf("Iter: %3d || Test Accuracy: %2.3f\n", iter, test_accuracy)
        # CUDA.reclaim()
        # CUDA.memory_status()
    end
end

# Train the NN-ODE and monitor the loss and weights.
Flux.@epochs 10 Flux.train!(loss, Flux.params(down, nn_ode.p, fc), CuIterator(train_dataloader), opt, cb=cb)
```

Overall, I am seeing ~11% GPU memory utilization reported from `nvidia-smi`. Definitely not OOM errors. In theory, I shouldn't have to load the data into main memory for `MNISTDataset()`, and instead, I should be able to lazy read from disk in the background. But I ran into some issues with that. I'll update if I work my way around those.

I am also not sure what to expect in terms of training time here (maybe @jessebett can comment). Also how many epochs is required? After 10 epochs, the test accuracy barely changes from ~10%.

darsnack:
> In DiffEqArray_to_Array we have solution dimensions [W,H,C,B,T] (right?). So we need to access via xarr[:,:,:,:,1] opposed to xarr[:,:,:,1,:]

Does this comment need to be addressed to fix convergence issues? (sorry I know nothing about neural ODEs beyond what they are)

darsnack:
By the way, I managed to get lazy loading all the way to disk working with this change for `MNISTDataset`:
```julia
struct MNISTDataset
end
LearnBase.nobs(::MNISTDataset) = 60_000
function LearnBase.getobs(::MNISTDataset, idx)
    imgs, labels = MNIST.traindata(Float32, idx)

    return reshape(imgs, size(imgs, 1), size(imgs, 2), 1), vec(onehot(array_wrap(labels)))
end
```

ChrisRackauckas:
@jessebett it would be great if you could take another stab with these modifications in mind.

DhairyaLGandhi:
You probably don't want to lazy load from disk for more real world cases, it would bring in IO latency in addition to the network io to move from RAM to VRAM. To me it seems like we aren't calling GC enough in the original case to trigger collection of allocated memory by CUDA.jl, this simply adds in the yield points to get GC to trigger. DataLoader isn't supposed to have GC points for the most part, but we should be able to suggest when it is safe to collect.

As a check, I was able to keep memory usage under 900MB and 2GB with varying frequencies of calling into the GC with this script.

```julia
using DiffEqFlux, OrdinaryDiffEq, Flux, Printf
using Flux.Losses: logitcrossentropy
using Flux.Data: DataLoader
using MLDatasets
using MLDataUtils:  LabelEnc, convertlabel, stratifiedobs
using CUDA
CUDA.allowscalar(false)

function loadmnist(batchsize = bs, train_split = 0.9)
    # Use MLDataUtils LabelEnc for natural onehot conversion
    onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw,
                                      LabelEnc.NativeLabels(collect(0:9)))
    # Load MNIST
    imgs, labels_raw = MNIST.traindata();
    # Process images into (H,W,C,BS) batches
    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
    y_data = onehot(labels_raw)
    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
                                                      p = train_split)
    return (
    # Use Flux's DataLoader to automatically minibatch and shuffle the data
       DataLoader((x_train, Float32.(y_train)); batchsize = batchsize,
                 shuffle = true),
    # Don't shuffle the test data
      DataLoader((x_test, Float32.(y_test)); batchsize = batchsize,
                shuffle = false)
     )
end

function main(bs = 128, train_split = 0.9)

  train_dataloader, test_dataloader = loadmnist(bs, train_split);
  T = [first(train_dataloader)]
  for i in train_dataloader
    push!(T, i)
  end
  gt = cu.(T)



  down = Chain(
               Conv((3,3),1=>64,relu), # GroupNorm(64,64),
               Conv((4,4),64=>64,relu,stride=(2,2),pad=(1,1)), # GroupNorm(64,64),
               Conv((4,4),64=>64,stride=(2,2),pad=(1,1)),
             ) |> gpu

  dudt = Chain(
               Conv((3,3),64=>64,relu,pad=(1,1)),
               Conv((3,3),64=>64,relu,pad=(1,1))
             ) |> gpu

  # fc = Chain(GroupNorm(64,64), x -> relu.(x),
  fc = Chain(  # x->relu.(x),
               MeanPool((6,6)),
               x -> reshape(x, 64,:),
               Dense(64,10)
          ) |> gpu

  nn_ode = NeuralODE(dudt, (0.f0, 1.f0), Tsit5(),
                     save_everystep = false,
                     reltol = 1e-3, abstol = 1e-3,
                     save_start = false) |> gpu

  model = Chain(
                down,             # (28,28,1,BS) -> (6,6,64,BS)
  #              nn_ode,           # (6,6,64,BS) -> (6,6,64,BS)
  #              DiffEqArray_to_Array,
                x -> reshape(x, 6,6, 64, :),
                fc                # (6,6,64,BS) -> (10, BS)
  )


  img, lab = gpu(train_dataloader.data[1][:, :, :, 1:1]), gpu(train_dataloader.data[2][:, 1:1])

  loss(x, y) = logitcrossentropy(model(x), y)
  # loss(x, y) = Flux.mse(model(x), y)

  opt = ADAM(0.001)
  iter = 0

  function cb()
    global iter += 1
    # Monitor that the weights do infact update
    # Every 10 training iterations show accuracy
    print(iter)
    print(" ")
    println(loss(img, lab))
  end

  ps = Flux.params(down, nn_ode.p, fc)
  for (i,d) in enumerate(gt)
    gs = gradient(ps) do
      loss(d...)
    end
    Flux.Optimise.update!(opt, ps, gs)
    if i % 10 == 0
      GC.gc()
      CUDA.reclaim()
    end
  end

  model
end
```


```
Thu Nov 26 01:56:34 2020                                                                                                 [345/956]
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 455.23.05    Driver Version: 455.23.05    CUDA Version: 11.1     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   1  Tesla P100-PCIE...  On   | 00000000:05:00.0 Off |                    0 |
| N/A   42C    P0    38W / 250W |    785MiB / 16280MiB |     32%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|    1   N/A  N/A   2140899      C   ...dhi/julia-1.5.2/bin/julia      761MiB |
+-----------------------------------------------------------------------------+
```

DhairyaLGandhi:
You could pass as a callback too

```julia
cb2 = Flux.throttle(() -> (GC.gc(); CUDA.reclaim()), 10)
```

darsnack:
Most real world ￼cases use lazy loading,￼ and it is the default in other frameworks. You don't pay the latency cost, since it happens in separate threads￼ and is buffered. The cost of moving to GPU memory is there in any system that doesn't move the data to the GPU all at once. I have been thinking about specializing `CuIterator` on our loader to keep two batches in memory so that even the GPU transfer cost is masked.

In my testing, the GC was able to keep the memory utilization well under 1GB with just `CuIterator` and no specific GC calls. On another note, I had to address @RobWalt comments to get the loss to meaningfully change. Seems there is still an issue with the LR though.

ToucheSir:
@DhairyaLGandhi how does GPU usage look with that periodic collection? My testing had it drop precipitously whenever `CUDA.reclaim()` was called. That may just be elimination of useless work, but if it is indeed affecting throughput then I don't think it's a great solution.

Zooming out a bit, one thing that dramatically slows down the full program is the per(-N)-step re-calculation of accuracy on all training and testing data. Ideally the model predictions would be passed to the callback and the metric would be incrementally calculated (either manually or using something like OnlineStats).

darsnack:
Yeah that's why I commented out the training line in my snippet. But in actuality it only does 100 batches of training data, so it isn't too bad. Using an incremental metric would be nicer. If manual GC/reclaim calls are required, then it should be part of `CuIterator`'s behavior (but I didn't see a need for manual calls).

DhairyaLGandhi:
Takes roughly the same amount of time as without, so I don't think its meaningfully affecting throughput in this situation.

DhairyaLGandhi:
@darsnack just ran your snippet - it seems to fill up the VRAM completely for me.

darsnack:
The differences are DataLoader and GC calls. Since DataLoader doesn't touch the GPU in either version of the code, the only thing could be the GC calls? It's weird...I ran your snippet and saw the same overall utilization.

DhairyaLGandhi:
> same overall utilization.

Could you clarify? Are you saying that the GPU utilisation was the same in my snippet as in the original?

ToucheSir:
My testing seemed to indicate that total VRAM usage (as reported by `nvidia-smi`, so including cached allocations) would _only_ decrease meaningfully after a `CUDA.reclaim()`. However, that nukes the allocator's pool, so I'm not sure that it's necessary if the model trains just as fast + doesn't OOM. Given how other models don't stress out VRAM as much, I still think the problem lies in the (DiffEq)Flux end instead of CUDA.jl.

DhairyaLGandhi:
Without yield points, CUDA.jl wants to allocate till it OOMs and then it reclaims the memory. The GC calls are meant to add those yield points in. You can try this by calling something like `f` a few times in succession and monitoring the GPU memory usage

```julia
function f(x)
  for _ = 1:10^4
    r = x .+ x
  end
end

x = rand(Float32, 28,28,1,128) |> gpu
```

darsnack:
> Could you clarify? Are you saying that the GPU memory utilisation was the same in my snippet as in the original?

No I meant that the GPU memory utilization was about the same in my snippet and your snippet. Which is not surprising since both snippets relieve GPU memory pressure by using `CuIterator`.

> CUDA.jl wants to allocate till it OOMs and then it reclaims the memory.

If you look at total VRAM allocation (reported by running `nvidia-smi`), then CUDA.jl will allocate almost all the available VRAM, but the volatile memory utilization remains < 7% (reported by `nvidia-smi -q -g 0 -d UTILIZATION`). Note that if I run something else on the GPU along with the program, *CUDA.jl only grabs the available memory* instead of the full 12GB available on my device (it doesn't OOM). Using `CUDA.reclaim()` reduces the total VRAM allocation to 1GB (like Brian mentioned), but it goes back up to 12GB within seconds. I don't think killing the allocation pool is doing much here for overall performance of the program.

Here is what I've found by running several variants.
| Snippet Name | DataLoader Variant        | Calls `GC.gc()`+`CUDA.reclaim()` | Epoch Runtime `@time` | GPU Utilization | Volatile GPU Memory Utilization | Max Allocated VRAM |
|:-------------|:--------------------------|:---------------------------------|----------------------:|----------------:|--------------------------------:|-------------------:|
| Snippet 1    | `Flux.Data.DataLoader`    | Yes                              | 557.514435 seconds    | ~40%            | ~7%                             | 12 GB / 12 GB      |
| Snippet 2    | `Flux.Data.DataLoader`    | No                               | 521.590281 seconds    | ~50%            | ~7%                             | 12 GB / 12 GB      |
| Snippet 3    | `DataLoaders.DataLoader`  | No                               | 484.777086 seconds    | ~50%            | ~7%                             | 12 GB / 12 GB      |
| Original     | `Flux.Data.DataLoader`    | No                               | 568.442654 seconds    | ~40%            | ~7%                             | 12 GB / 12 GB      |

[Here is a gist](https://gist.github.com/darsnack/28289e94e267b62ca04bb990567ae37a) for all the variants that I ran. I killed the Julia process completely between each test to release all memory on the GPU.

Seems like I see virtually no difference between these variants. Maybe someone else can check, but all of them allocate 12GB for me. Snippet 1 will drop to ~2GB when `CUDA.reclaim()` is called, but the effect of that lasts only for ~2 seconds before jumping back to 12 GB. It's weird to me that the Original variant doesn't have different memory stats. It clearly loads the full dataset into GPU memory in one go.

ToucheSir:
Turns out I had some REPL state hanging around. `CUDA.reclaim()` (no `GC.gc()`, I assume the former also introduces a yield point?) is required just once before training, but only if the accuracy has been "burned in". I was only able to repro `snippet-5.jl` though, snippet 3 complains about a missing `iterate` method for the DataLoader wrappers?


> It's weird to me that the Original variant doesn't have different memory stats. It clearly loads the full dataset into GPU memory in one go.

Some back-of-the-napkin math suggests that the entire MNIST dataset only occupies ~600MB. Usage only ramps up once one actually calls the model.

It's odd that using `CUDA.gc(false)` as per https://github.com/JuliaGPU/CUDA.jl/issues/137 doesn't seem to change much. Similar story with the split pool, though I'm not sure if I was using it properly. Could the OOMs be related to https://github.com/JuliaGPU/CUDA.jl/issues/340?

darsnack:
> Could the OOMs be related to JuliaGPU/CUDA.jl#340?

Not sure about how related it is, but I definitely faced a similar error (bad status, nothing pointer) when running tests.

> snippet 3 complains about a missing iterate method for the DataLoader wrappers?

You have DataLoaders.jl installed and closed the REPL between runs?

darsnack:
> CUDA.reclaim() (no GC.gc(), I assume the former also introduces a yield point?) is required just once before training, but only if the accuracy has been "burned in".

I'll have to try this. You didn't see the allocation jump back up? I was seeing it drop every time the callback ran then jump a few seconds later.

FYI, I'm running Julia 1.5.0.

DhairyaLGandhi:
I don't see the behaviour of jumping memory back up. It of course increases, but doesn't go on to consume all the VRAM (16GB in my case). As mentioned, maybe you want to use https://github.com/SciML/DiffEqFlux.jl/pull/387#issuecomment-734156816 as a callback. I would perhaps remove the calls to accuracy and the logging from the training cost, these should be separate anyway.

Also, you might want to put things into a `main` function and avoid globals. Lastly, it still counts compilation time for a bunch of things. You might want to get rid of that too?

I barely see `nvidia-smi` reporting anything over 4-4.5GB.

darsnack:
Okay I'll try with those changes.

ChrisRackauckas:
So what's the final say here? @avik-pal do you have comments on how we finish this? I would like to just get something usable into the docs.

avik-pal:
@ChrisRackauckas I confirmed that we can train this without running into memory issues with Cuda3.0
```julia
using DiffEqFlux, OrdinaryDiffEq, Flux, Printf
using Flux: logitcrossentropy
using Flux.Data: DataLoader
using MLDatasets
using MLDataUtils:  LabelEnc, convertlabel, stratifiedobs
using CUDA
CUDA.allowscalar(false)

function loadmnist(batchsize = bs, train_split = 0.9)
    # Use MLDataUtils LabelEnc for natural onehot conversion
    onehot(labels_raw) = convertlabel(LabelEnc.OneOfK, labels_raw,
                                      LabelEnc.NativeLabels(collect(0:9)))
    # Load MNIST
    imgs, labels_raw = MNIST.traindata();
    # Process images into (H,W,C,BS) batches
    x_data = Float32.(reshape(imgs, size(imgs,1), size(imgs,2), 1, size(imgs,3)))
    y_data = onehot(labels_raw)
    (x_train, y_train), (x_test, y_test) = stratifiedobs((x_data, y_data),
                                                         p = train_split)
    return (
        # Use Flux's DataLoader to automatically minibatch and shuffle the data
        DataLoader(gpu.(collect.((x_train, y_train))); batchsize = batchsize,
                   shuffle = true),
        # Don't shuffle the test data
        DataLoader(gpu.(collect.((x_test, y_test))); batchsize = batchsize,
                   shuffle = false)
    )
end

# Main
const bs = 128
const train_split = 0.9
train_dataloader, test_dataloader = loadmnist(bs, train_split);

down = Chain(
             Conv((3,3),1=>64,relu,stride=1), GroupNorm(64,64),
             Conv((4,4),64=>64,relu,stride=2,pad=1), GroupNorm(64,64),
             Conv((4,4),64=>64,stride=2,pad=1),
            )|>gpu;
dudt = Chain(
           Conv((3,3),64=>64,tanh,stride=1,pad=1),
           Conv((3,3),64=>64,tanh,stride=1,pad=1)
          ) |>gpu;
          
fc = Chain(GroupNorm(64,64),
           x->relu.(x),
           MeanPool((6,6)),
           x -> reshape(x,(64,:)),
           Dense(64,10)
          )|>gpu;
          
nn_ode = NeuralODE(dudt, (0.f0, 1.f0), Tsit5(),
                   save_everystep = false,
                   reltol = 1e-3, abstol = 1e-3,
                   save_start = false) |> gpu


function DiffEqArray_to_Array(x)
    xarr = gpu(x)
    return xarr[:,:,:,:,1]
end

# Build our over-all model topology
model = Chain(
              down,             #(28,28,1,BS) -> (6,6,64,BS)
              nn_ode,           #(6,6,64,BS) -> (6,6,64,BS)
              DiffEqArray_to_Array,
              fc                #(6,6,64,BS) -> (10, BS)
             )

# To understand the intermediate NN-ODE layer, we can examine it's dimensionality
img, lab = train_dataloader.data[1][:, :, :, 1:1], train_dataloader.data[2][:, 1:1]

x_d = down(img)

# We can see that we can compute the forward pass through the NN topology
# featuring an NNODE layer.
x_m = model(img)

classify(x) = argmax.(eachcol(x))

function accuracy(model, data; n_batches = 10)
    total_correct = 0
    total = 0
    for (i, (x, y)) in enumerate(data)
        # Only evaluate accuracy for n_batches
        i > n_batches && break
        target_class = classify(cpu(y))
        predicted_class = classify(cpu(model(x)))
        total_correct += sum(target_class .== predicted_class)
        total += length(target_class)
    end
    return total_correct / total
end

# burn in accuracy
accuracy(model, train_dataloader)

loss(x, y) = logitcrossentropy(model(x), y)

# burn in loss
loss(img, lab)

opt = ADAM(0.05)
iter = 0

cb() = begin
    global iter += 1
    # Monitor that the weights do infact update
    # Every 10 training iterations show accuracy
    if iter % 10 == 1
        train_accuracy = accuracy(model, train_dataloader) * 100
        test_accuracy = accuracy(model, test_dataloader;
                                 n_batches = length(test_dataloader)) * 100
        @printf("Iter: %3d || Train Accuracy: %2.3f || Test Accuracy: %2.3f\n",
                iter, train_accuracy, test_accuracy)
    end
end

Flux.@epochs 10 Flux.train!(loss, Flux.params(down, nn_ode.p, fc), train_dataloader, opt, cb = cb)
```
Training with batch size = 512 requires about 32 GB mem, but I could train it on my local 4GB GPU at batch size ~ 100

DhairyaLGandhi:
This is promising. Maybe we can tune the memory needs by not loading the whole dataset into the memory at once and have it be handled in a different process? I'm not sure if it would matter too much for mnist, but considering memory limitations, it might be crucial for real work loads

DhairyaLGandhi:
@avik-pal could you also try with removing the logging? It could help to benchmark the two as well

avik-pal:
@DhairyaLGandhi u mean benchmark the CUDA3.0 version with the older versions?

DhairyaLGandhi:
I meant benchmark with and without logging. Apologies, should have been clearer.

avik-pal:
Even without logging the memory usage seems to hit the max available GPU memory.
> EDIT: The problem with occupying the entire GPU memory still persists though. bs = 128 takes 3 GB locally but on cyclops it occupies 31GB. Though I feel that is a issue best handled elsewhere and get this merged to provide a tutorial for users.

DhairyaLGandhi:
I agree we should merge this but wanted to note any todos from here.

ChrisRackauckas:
Yes. I guess rebase or make a new PR and let's get this merged into the docs. I would also like to have an integration test here to check for any regressions (hopefully our GPU CI machines can handle it!)

avik-pal:
I will go ahead and make a new PR.

ChrisRackauckas:
Thanks @jessebett for getting us started here!

darsnack:
@avik-pal You could try DataLoaders.jl which does load the dataset as needed on a separate process as @DhairyaLGandhi suggested.

ChrisRackauckas:
For everyone following this, I think just the CUDA update alone fixed it:

>1 epoch with 512 batch size takes 22s with CUDA3.0 while the released version of Flux takes around 80-90s. The numbers are with logging removed

https://github.com/SciML/DiffEqFlux.jl/pull/532#issuecomment-821111929

PR #410, opened 2020-09-09
Update TagBot.yml

kanav99:
@ChrisRackauckas this should fix the tags issue, we will get tags from the next releases.


ChrisRackauckas:
oh nice!

PR #412, opened 2020-09-09
Fix parameters for `NeuralODE{TensorLayer}`

### LINKED ISSUE #411: `NeuralODE{TensorLayer}` has no parameters
Not sure if this is a bug but I'm using DiffEqFlux v1.22.0 and noticed that

```julia
using OrdinaryDiffEq, Flux, DiffEqFlux
NN = TensorLayer([LegendreBasis(4), LegendreBasis(4)], 4)
node = NeuralODE(NN, (0.0, 1.0), ROCK4())

julia> NN.p |> length
64

julia> node.p
0-element Array{Any,1}
```

so if you try `sciml_train` on `node` you get an error:

```julia
julia> DiffEqFlux.sciml_train(nothing, node.p, Descent(), maxiters=1)
ERROR: MethodError: no method matching typemax(::Type{Any})
Closest candidates are:
  typemax(::Type{Bool}) at bool.jl:7
  typemax(::Type{Int8}) at int.jl:652
  typemax(::Type{UInt8}) at int.jl:654
  ...
Stacktrace:
 [1] sciml_train(::Nothing, ::Array{Any,1}, ::Descent, ::Base.Iterators.Cycle{Tuple{DiffEqFlux.NullData}}; cb::Function, maxiters::Int64, progress::Bool, save_best::Bool) at /home/alir/.julia/packages/DiffEqFlux/1qqym/src/train.jl:114
 [2] top-level scope at REPL[33]:1
```

Should work if you pass in `NN.p` instead of `node.p` but seems like a bug as you usually pass in the parameters of the `NeuralODE`.

ali-ramadhan:
I added a new constructor for `NeuralODE(model::TensorLayer, ...)` which required me to `include` the `tensor_product_*.jl` files first.

Now `NeuralODE{TensorLayer).p` has parameters as expected:

```julia
julia> using Test, OrdinaryDiffEq, Flux, DiffEqFlux

julia> NN = TensorLayer([LegendreBasis(4), LegendreBasis(4)], 4)

julia> node = NeuralODE(NN, (0.0, 1.0), ROCK4())

julia> node.p |> length
64

julia> @test all(node.p == NN.p)
Test Passed
```

Resolves #411

### DIFF
--- a/src/DiffEqFlux.jl
+++ b/src/DiffEqFlux.jl
@@ -75,11 +75,11 @@ ZygoteRules.@adjoint Tridiagonal(dl, d, du) = Tridiagonal(dl, d, du), p̄ -> (di
 include("ffjord.jl")
 include("train.jl")
 include("fast_layers.jl")
+include("tensor_product_basis.jl")
+include("tensor_product_layer.jl")
 include("neural_de.jl")
 include("require.jl")
 include("spline_layer.jl")
-include("tensor_product_basis.jl")
-include("tensor_product_layer.jl")
 include("collocation.jl")
 include("hnn.jl")
 
--- a/src/neural_de.jl
+++ b/src/neural_de.jl
@@ -76,7 +76,7 @@ function (n::NeuralODE)(x,p=n.p)
     solve(prob,n.args...;sense=sense,n.kwargs...)
 end
 
-function (n::NeuralODE{M})(x,p=n.p) where {M<:FastChain}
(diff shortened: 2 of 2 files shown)


ali-ramadhan:
Ah after a discussion with @ChrisRackauckas on Slack I think I misunderstood what `TensorLayer` does. Please close this PR if #411 is not an actual bug.

PR #415, opened 2020-09-11
Batched Implementation for FFJORD

### LINKED ISSUE #342: FFJORD for batched data
I looked into the FFJORD [implementation](https://github.com/SciML/DiffEqFlux.jl/blob/8cb89e490eb90fa74edcde363386b4799c1bfb95/src/ffjord.jl) and [tests](https://github.com/SciML/DiffEqFlux.jl/blob/8cb89e490eb90fa74edcde363386b4799c1bfb95/test/cnf_test.jl) and noticed that it does not seem to work with batched training. I started looking into how to do this, but since it is not done in this implementation perhaps the problem is harder than I realize. Could someone shed some light on the situation? @d-netto perhaps?

### LINKED ISSUE #413: CNF layers very slow with moderately sized Neural Networks
Hello, I'm running into a performance issue when using CNF layers for a simple problem. In this code (it's basically the tutorial with a larger network)

```julia
using DiffEqFlux
using OrdinaryDiffEq
using Flux
using Distributions
using Zygote

nn = Chain(Dense(1, 50, tanh),
           Dense(50, 50, tanh),
           Dense(50, 1, tanh))

tspan = (0.0,10.0)
ffjord_test = FFJORD(nn, tspan, Tsit5())

data_train = [Float32(rand(Normal(6.0,0.7))) for i in 1:100]

function loss_adjoint(θ)
    logpx = [ffjord_test(x,θ) for x in data_train]
    loss = -mean(logpx)
end

function train_callback(θ, l)
    println(l)
    false
end

res1 = DiffEqFlux.sciml_train(loss_adjoint,
                              ffjord_test.p,
                              ADAM(0.1),
                              cb = train_callback,
                              maxiters = 10)
```
the first print of the loss happens for me after about 5mins, and it doesn't really speed up after that. I would have expected much faster training times.

I'm using  Julia 1.5.0 and
Zygote v0.5.6
DiffEqFlux v1.22.0
OrdinaryDiffEq v5.42.7
Flux v0.11.1
Distributions v0.23.11

avik-pal:
TODOs:
- [x] Test that this works ~(this was part of another codebase and I haven't yet tested if it integrates correctly here)~
- [ ] Type Inference fails
- [x] Batched version for the non-monte carlo solution.

Fixes #413
Fixes #342 

Some runtime stats:
For a Batch size of 100 on CPU, it takes 2.2 mins per iteration of training.
For a Batch size of 1k on GPU, it takes 0.52s per training iteration.

All of these use the model from #413

### DIFF
--- a/docs/src/examples/normalizing_flows.md
+++ b/docs/src/examples/normalizing_flows.md
@@ -8,7 +8,7 @@ We can use DiffEqFlux.jl to define, train and output the densities computed by C
 using DiffEqFlux, OrdinaryDiffEq, Flux, Optim, Distributions, Zygote
 
 nn = Chain(Dense(1, 3, tanh), Dense(3, 1, tanh))
-tspan = (0.0,10.0)
+tspan = (0.0f0,10.0f0)
 ffjord_test = FFJORD(nn,tspan, Tsit5())
 ```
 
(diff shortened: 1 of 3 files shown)


avik-pal:
DeterministicCNF is now deprecated. We can remove it in a later release.

ChrisRackauckas:
It would probably be good to add a GPU test for FFJORD, and answer that issue with how the performance has changed. It should be much better now.

PR #416, opened 2020-09-11
work around trace for neldermead

ChrisRackauckas:
@pkofod is there a list of these somewhere? Or is this the only one?


Vaibhavdixit02:
This was fixed here btw https://github.com/SciML/GalacticOptim.jl/blob/master/src/solve.jl#L152

PR #421, opened 2020-09-27
Add GalacticOptim and update tests to use it

Vaibhavdixit02:
This PR changes the tests in layers_sciml.jl to use GalacticOptim to phase out `sciml_train`


ChrisRackauckas:
There's a lot more `sciml_train` calls than that, right?

ChrisRackauckas:
I guess we can merge this for now as a start, but what I meant was replacing all `sciml_train` calls and deprecating it.

Vaibhavdixit02:
I guess I should have added WIP to the title 😂

PR #424, opened 2020-09-28
Add warning to sciml_train, add a dispatch of sciml_train to GalacticOptim and rename layer_sciml test file to galacticoptim 

Vaibhavdixit02:
The PR adds a deprecation warning for `sciml_train` and updates tests to use GalacticOptim


src/DiffEqFlux.jl
```diff
@@ -27,6 +27,11 @@ function diffeq_adjoint(p,prob,solver=nothing,args...;u0=prob.u0,kwargs...)
   concrete_solve(prob,solver,u0,p,args...;kwargs...)
 end
 
+function sciml_train(loss,θ,opt, args...; kwargs...)
+  @warn("sciml_train has been deprecated in favor of GalacticOptim.jl (https://github.com/SciML/GalacticOptim.jl)")
+  sciml_train(loss,θ,opt, args...; kwargs...)
```
ChrisRackauckas, line 32: that's going to be an infinite loop?

test/cnf_test.jl
```diff
@@ -32,7 +32,9 @@ function loss_adjoint(θ)
     loss = -mean(logpx)
 end
 
-res = DiffEqFlux.sciml_train(loss_adjoint, cnf_test.p,
+optfunc = GalacticOptim.OptimizationFunction((x, p) -> loss_adjoint(x), cnf_test.p, GalacticOptim.AutoForwardDiff())
```
ChrisRackauckas, line 35: Why are these not using Zygote?

test/neural_ode_mm.jl
```diff
@@ -37,5 +37,7 @@ cb = function (p,l,pred) #callback function to observe training
 end
 
 l1 = first(loss(ndae.p))
-res = DiffEqFlux.sciml_train(loss, ndae.p, BFGS(initial_stepnorm = 0.001), cb = cb, maxiters = 100)
+optfunc = GalacticOptim.OptimizationFunction((x, p) -> loss(x), ndae.p, GalacticOptim.AutoForwardDiff())
```
ChrisRackauckas, line 40: why not Zygote?

test/newton_neural_ode.jl
```diff
@@ -23,11 +23,14 @@ nODE = NeuralODE(NN, tspan, ROCK4(), reltol=1e-4, saveat=[tspan[end]])
 
 loss_function(θ) = Flux.mse(y, nODE(x, θ))
 l1 = loss_function(nODE.p)
-res = DiffEqFlux.sciml_train(loss_function, nODE.p, LBFGS(), maxiters = 400, cb=cb)
+
+optfunc = GalacticOptim.OptimizationFunction((x, p) -> loss_function(x), nODE.p, GalacticOptim.AutoForwardDiff())
```
ChrisRackauckas, line 27: why not Zygote?

ChrisRackauckas:
Zygote should be the default differentiation method here.

src/train.jl
```diff
@@ -3,6 +3,8 @@ const DEFAULT_DATA = Iterators.cycle((NullData(),))
 Base.iterate(::NullData, i=1) = nothing
 Base.length(::NullData) = 0
 
+@deprecate sciml_train(loss, θ, opt, args...; kwargs...) GalacticOptim.solve(optprob, opt)
```
ChrisRackauckas, line 6: this deprecation is going to fail since `optprob` is not defined.

Vaibhavdixit02: Not sure how this deprecation should be handled actually. Can you be more explicit?

ChrisRackauckas: Make it be a function call that builds the standard GalacticOptim problem and calls solve on it, but in there throws a warning that the function is deprecated. I think it's slightly too complicated for the standard `@deprecate` mechanism, though if you can shove it all into one call for that it would be nice.

Vaibhavdixit02: Hmm, can we just remove `sciml_train` now? (Easiest way I guess, otherwise I'll try the above)

ChrisRackauckas: No.

src/train.jl
```diff
@@ -3,7 +3,12 @@ const DEFAULT_DATA = Iterators.cycle((NullData(),))
 Base.iterate(::NullData, i=1) = nothing
 Base.length(::NullData) = 0
 
-@deprecate sciml_train(loss, θ, opt, args...; kwargs...) GalacticOptim.solve(optprob, opt)
+function sciml_train(loss, θ, opt, adtype; kwargs...)
+  @warn("sciml_train has been deprecated in favor of GalacticOptim.jl (https://github.com/SciML/GalacticOptim.jl)")
```
Vaibhavdixit02, line 7: Should this warning be added to all the `sciml_train` calls?

Vaibhavdixit02: @ChrisRackauckas?

ChrisRackauckas: yeah that would be good.

Vaibhavdixit02:
I think this should be good to go in once tests pass

src/train.jl
```diff
@@ -1,404 +1,6 @@
...
-                                        false, #f_increased,
-                                        nothing,
-                                        maxiters,
-                                        maxiters,
-                                        0,
-                                        true,
-                                        NaN,
-                                        bboptre.elapsed_time,
-                                        NamedTuple())
+function sciml_train(loss, θ, opt, adtype::DiffEqBase.AbstractADType = GalacticOptim.AutoZygote(); kwargs...)
```
ChrisRackauckas, line 1: this doesn't handle the data iterator?

ChrisRackauckas: Replacing the 4th argument with the data iterator will likely cause many cases to fail, like the MNIST one. Instead, pass the data iterator to GalacticOptim (it has a data iterator... right?) and just implicitly hardcode this to AutoZygote (which is essentially what sciml_train had done before).

Vaibhavdixit02: GalacticOptim doesn't have a data iterator, I'll take a look at the mnist example and make the fixes

test/cnf_test.jl
```diff
@@ -111,9 +106,7 @@ function loss_adjoint(θ)
     return mean(@. -logpx + 0.1 * λ₁ + λ₂)
 end
 
-res = DiffEqFlux.sciml_train(loss_adjoint, 0.01f0 .* ffjord_test.p,
-                             ADAM(0.1), cb = cb,
-                             maxiters = 100)
+res = DiffEqFlux.sciml_train(loss_adjoint, 0.01f0 .* ffjord_test.p, ADAM(0.1), GalacticOptim.AutoZygote(), cb=cb, maxiters = 300)
```
ChrisRackauckas, line 109: after making sure tests pass, all of these sciml_train calls should be replaced with GalacticOptim code. The tests should all use what we say is the correct way to do it.

Vaibhavdixit02: So the GalacticOptim way of doing it has been added in galacticoptim.jl tests, I wanted to (and that's what has been discussed here and in previous PRs) to keep around `sciml_train` that passes it forward to GalacticOptim (with a warning) and then once we decide to completely drop support for it I think we can switch all the tests to the GalacticOptim interface.

P.S. The `sciml_train` wrapper for GalacticOptim is very user friendly

ChrisRackauckas: >So the GalacticOptim way of doing it has been added in galacticoptim.jl tests, I wanted to (and that's what has been discussed here and in previous PRs) to keep around sciml_train that passes it forward to GalacticOptim (with a warning) and then once we decide to completely drop support for it I think we can switch all the tests to the GalacticOptim interface.

Yes, there should be a depwarn test, but we should have like one. Most of the tests should do what the tutorial does. (and the docs need an update, but that can come in a follow up)

test/newton_neural_ode.jl
```diff
@@ -39,11 +39,11 @@ nODE = NeuralODE(NN, tspan, ROCK2(), reltol=1e-4, saveat=[tspan[end]])
 
 loss_function(θ) = Flux.mse(y, nODE(x, θ))
 l1 = loss_function(nODE.p)
-res = DiffEqFlux.sciml_train(loss_function, nODE.p, LBFGS(), GalacticOptim.AutoZygote(), maxiters = 400, cb=cb, allow_f_increases = true)
+res = DiffEqFlux.sciml_train(loss_function, nODE.p, LBFGS(), GalacticOptim.AutoZygote(), maxiters = 1000, cb=cb)
 @test 10loss_function(res.minimizer) < l1
-res = DiffEqFlux.sciml_train(loss_function, nODE.p, NewtonTrustRegion(), GalacticOptim.AutoZygote(), maxiters = 400, cb=cb, allow_f_increases = true)
+res = DiffEqFlux.sciml_train(loss_function, nODE.p, NewtonTrustRegion(), GalacticOptim.AutoReverseDiff(), maxiters = 400, cb=cb)
```
ChrisRackauckas, line 44: That's an issue to solve. The Newton doesn't work with the adjoints anymore?

Vaibhavdixit02: It does, it's just very slow. I was surprised that changing the AD backend did it, It think the issue is in `FastChain`

Vaibhavdixit02: Actually it's an issue when using `ROCK2` with `FastChain` `ROCK4`

Vaibhavdixit02: And this test has been timing out on some runs from before https://travis-ci.com/github/SciML/DiffEqFlux.jl/builds/189598780

Vaibhavdixit02:
@ChrisRackauckas I think this has become quite big already and the aim of the PR imo has been achieved, the subsequent change of all the tests to GalacticOptim interface should happen later since even now (until all docs and tutorials are updated) users are going to stick to `sciml_train` and having these tests use it would be better till we are ready to completely remove the `sciml_train` function itself?

ChrisRackauckas:
>users are going to stick to sciml_train and having these tests use it would be better till we are ready to completely remove the sciml_train function itself?

Users are going to switch once we start throwing warnings telling them to switch. So either we don't do the warning for now, or we switch the syntax.

ChrisRackauckas:
Cheers!

ChrisRackauckas:
Time to do the documentation updates.

PR #455, opened 2020-12-16
Create tutorial for Bayesian Neural ODE: NUTS. 

RajDandekar:


docs/src/examples/BayesianNODE_NUTS.md
```diff
@@ -0,0 +1,254 @@
...
+samples_reduced = samples[1:5, :]
+
+samples_reshape = reshape(samples_reduced, (500, 5, 1))
+
+Chain_Spiral = Chains(samples_reshape)
+
+plot(Chain_Spiral)
+
+autocorplot(Chain_Spiral)
+```
```
Vaibhavdixit02, line 114: Add the images of the plots here, take a look at the other md files with images for how to do it if you are unfamiliar

RajDandekar: Yup. Added plots of chain diagnosis as well as the retrodicted plots.

docs/src/examples/BayesianNODE_NUTS.md
```diff
@@ -0,0 +1,277 @@
+# Bayesian Neural ODEs: NUTS
+The DiffEqFlux.jl library in Julia can be seamlessly combined with Bayesian estimation
+librarries likes AdvancedHMC.jl and Turing.jl. This enables converting Neural ODEs
```
Vaibhavdixit02, line 3:
```suggestion
libraries likes AdvancedHMC.jl and Turing.jl. This enables converting Neural ODEs
```

Vaibhavdixit02: I think we need a more convincing introduction, this doesn't show the motivation of why we want to integrate bayesian with neural odes, just throw some of the abstract from the paper here

docs/src/examples/BayesianNODE_NUTS.md
```diff
@@ -0,0 +1,277 @@
...
+Auto-Corelation Plot:
+
+![](https://github.com/RajDandekar/BayesianNeuralODE.jl/blob/master/BayesianNODE_NUTS_ChainPlot2.png?raw=true)
+
+
+
+
+## Explanation
+
+Step1:Get the data from the Spiral ODE example
```
Vaibhavdixit02, line 142:
```suggestion
#### Step 1: Get the data from the Spiral ODE example
```

Vaibhavdixit02: And similarly for the other steps, I think it might looks better nothing more 😅

docs/src/examples/BayesianNODE_NUTS.md
```diff
@@ -0,0 +1,277 @@
...
+function trueODEfunc(du, u, p, t)
+    true_A = [-0.1 2.0; -2.0 -0.1]
+    du .= ((u.^3)'true_A)'
+end
+
+prob_trueode = ODEProblem(trueODEfunc, u0, tspan)
+ode_data = Array(solve(prob_trueode, Tsit5(), saveat = tsteps))
+```
+
+Step2: Define the Neural ODE architecture. Note that this step potentially offers a lot of flexibility in the number of layers/ number of units in each layer. It may not necessarily be true that a 100 units
```
Vaibhavdixit02, line 159: Split out the step heading and the description, in step 4 as well

docs/src/examples/BayesianNODE_NUTS.md
RajDandekar: It's good to focus on the asthetics..Sometimes, I don't focus too much on that, appreciate the suggestion.

docs/src/examples/BayesianNODE_NUTS.md
RajDandekar: My thinking was that when a person comes to visit this tutorial,  having seen the paper, that person would be convinced of why this integration is needed.

However, doesn't hurt to be more comprehensive. 

I have tried to make it comprehensive now.

docs/src/examples/BayesianNODE_NUTS.md
```diff
@@ -0,0 +1,278 @@
+# Bayesian Neural ODEs: NUTS
```
ChrisRackauckas, line 1: why only NUTS?

RajDandekar: In this tutorial, only NUTS is shown. 

For SGLD, we can put it in another tutorial. Else, I feel this tutorial will be too big.

ChrisRackauckas:
Follow up with the SGLD one.

PR #471, opened 2020-12-26
Scimltraindoc

artkuo:
Here's some documentation for `sciml_train`, in the form of a brief docstring and a small update to the markdown documentation. The latter briefly describes the struct that `sciml_train` returns. I realize that everything is transient as `GalacticOptim` is being developed, but I think there are others like me who are puzzling over how to call this function. Perhaps this could serve as a useful placeholder for the eventual form everything will take. Thanks for putting together an extremely impressive ecosystem in DiffEqFlux.


ChrisRackauckas:
Thanks! I'll merge for now as this is already an improvement, but note that this library has already switched its internals to GalacticOptim. We haven't documented that yet though, but we should really be removing sciml_train ASAP. That said, there's a ton of tutorials to update to do that though!

Issue #480, opened 2021-01-20
FastChain-Compatible Layers without bias term


brendankeith:
I would like to create a FastDense or StaticDense layer without a bias term. Something like this can be done in pytorch using a boolean flag (https://pytorch.org/docs/stable/generated/torch.nn.Linear.html#torch.nn.Linear) is this also possible with DiffEqFlux?

ChrisRackauckas:
It's not implemented yet but it's not more than an hour's work. I think this is a good feature to point new devs to.

aussetg:
#487

PR #482, opened 2021-02-03
FastChain-Compatible Layers without bias term

Aman027:
#### Reference  Issues/PR 
#480 
#### What does this implement/fix ?
These change add the functionality of keeping a bias term as an option while using FastDense or StaticDense. One can use
 bias = false for not keeping the bias term. The default value for bias is true and giving bias as an argument is optional.


src/fast_layers.jl
```diff
@@ -47,18 +48,28 @@ struct FastDense{F,F2} <: FastLayer
...
     W = reshape(@view(p[1:(f.out*f.in)]),f.out,f.in)
   end
 
-  b = ((f.bias == true) ? (p[(f.out*f.in+1):end]) : 0.0)
-  r = W*x .+ b
-  ifgpufree(b)
-
+  if f.bias == true
+    b = p[(f.out*f.in+1):end]
+    r = W*x .+b
```
ChrisRackauckas, line 68:
```suggestion
    r = W*x .+ b
```

src/fast_layers.jl
```diff
@@ -47,18 +48,28 @@ struct FastDense{F,F2} <: FastLayer
...
 # (f::FastDense)(x,p) = f.σ.(reshape(uview(p,1:(f.out*f.in)),f.out,f.in)*x .+ uview(p,(f.out*f.in+1):lastindex(p)))
-(f::FastDense)(x,p) = f.σ.(reshape(p[1:(f.out*f.in)],f.out,f.in)*x .+ ((f.bias == true) ? (p[(f.out*f.in+1):end]) : 0.0))
+function (f::FastDense)(x,p) 
+  y = f.σ.(reshape(p[1:(f.out*f.in)],f.out,f.in)*x)
+  if (f.bias == true)
+    return y .+ p[(f.out*f.in+1):end]
+  else
+    return y
+  end
+end
```
ChrisRackauckas, line 58:
```suggestion
end

```

src/fast_layers.jl
```diff
@@ -109,37 +126,52 @@ The `activation` function defaults to `identity`, meaning the layer is an affine
...
+    initial_params() = vcat(vec(initW(out, in)),initb(out))
+    new{out,in,bias,typeof(σ),typeof(initial_params)}(σ,initial_params)
   end
 end
 
 function param2Wb(f::StaticDense{out,in}, p) where {out,in}
-  _W = @views p[1:(out*in)]
-  _b =  ((f.bias == true ) ? (p[(out*in+1):end]) : 0 )
+  if f.bias == true
+  _W, _b = @views p[1:(out*in)], p[(out*in+1):end]
```
ChrisRackauckas, line 148: whitespace.

src/fast_layers.jl
```diff
@@ -109,37 +126,52 @@ The `activation` function defaults to `identity`, meaning the layer is an affine
...
+    W, b = param2Wb(f, p)
+    return f.σ.(W*x .+ b)
+  else 
+    W = param2Wb(f,p)
+    return f.σ.(W*x)
+  end
 end
 ZygoteRules.@adjoint function (f::StaticDense{out,in})(x,p) where {out,in}
+  if f.bias == true
   W, b = param2Wb(f, p)
```
ChrisRackauckas, line 169: whitespace

Project.toml
```diff
@@ -18,12 +18,14 @@ GalacticOptim = "a75be94c-b780-496d-a8a9-0878b188d577"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 LoggingExtras = "e6f89c97-d47a-5376-807f-9c37f3926c36"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
ChrisRackauckas, line 21: Plots should not be a dependency.

ChrisRackauckas, line 21:
```suggestion
```

Project.toml
```diff
@@ -18,12 +18,14 @@ GalacticOptim = "a75be94c-b780-496d-a8a9-0878b188d577"
...
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
 RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 TerminalLoggers = "5d786b92-1e48-4d6f-9151-6b4477ca9bed"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
ChrisRackauckas, line 28: Test should not be a dependency.

ChrisRackauckas, line 28:
```suggestion
```

ChrisRackauckas:
That's a failure on master, not your issue.

src/fast_layers.jl
```diff
@@ -96,9 +104,8 @@ ZygoteRules.@adjoint function (f::FastDense)(x,p)
   end
   y,FastDense_adjoint
 end
-paramlength(f::FastDense) = f.out*(f.in + 1)
+paramlength(f::FastDense) = length(f.initial_params())
```
ChrisRackauckas, line 107: revert and fix this: generating a whole parameter vector to check the length is very inefficient.

ChrisRackauckas, line 107:
```suggestion
paramlength(f::FastDense) = f.out*(f.in + f.bias)

```

src/fast_layers.jl
```diff
@@ -108,34 +115,53 @@ The `activation` function defaults to `identity`, meaning the layer is an affine
...
-  b = @inbounds SVector{out}(_b)
+  if f.bias == true
+    _W, _b = @views p[1:(out*in)], p[(out*in+1):end]
+    W = @inbounds convert(SMatrix{out,in},_W)
+    b = @inbounds SVector{out}(_b)
   return W, b
+  else 
+    W = @view p[1:(out*in)]
+    _W = @inbounds convert(SMatrix{out,in},W)
+    return _W
```
ChrisRackauckas, line 145: this now isn't type-stable. It should create `b = SVector{out}()` which is just empty.

Aman027: I didn't quite get your point. Why should we make and return `b = SVector{out}()` if we are not going to use it in the subsequent functions and as far as I know function returns tuple when f.bias == true and it is handled appropriately wherever it is used and same is the case when f.bias == false. The current error is in mnist_gpu.jl which fails even when I am using master repo of DiffEqFlux.jl.

ChrisRackauckas: It's fixed on master, but that's not the point. The point is speed. This code isn't type-stable because it's dependent on runtime information. The right thing to do might be to make `bias` a type parameter in the case of StaticDense to compensate, or just keep type-stability the easy way.

ChrisRackauckas:
oh can you do a PR to update the docs?

Aman027:
> oh can you do a PR to update the docs?

Sure, I can do that.

PR #490, opened 2021-02-14
updated minibatch example to use Flux call

khannay:
This is my first pull request. The point here is to update the mini batch example which currently fails at the sciml train call. I have rewritten the sample example just changing the syntax around so that it now runs.


ChrisRackauckas:
Thanks looks great! We should also find a nice way to link it to a test to track it over time. I think our only testing of the minibatching is implicitly through the MNIST tests.

PR #491, opened 2021-02-15
Controlling an SDE 

frankschae:
Adds a NSDE control example using continuous adjoint sensitivity methods.

Should we add the plot at the end which the code produces?
fidelity

Will the `ψ(t)` `dW_t `  be properly displayed?


ChrisRackauckas:
Awesome! I added the plot. I think it should all display. One way to find out.

PR #500, opened 2021-03-05
Updated sciml_train information.

mkg33:
I updated sciml_train information and deleted the redundant parts from require.jl.


ChrisRackauckas:
The goal here should be to remove `sciml_train` from all of the tutorial pages as well, and the tests. I think we should completely move it all to GalacticOptim.jl

mkg33:
All right: I thought we wanted to replace `sciml_train` with Galactic and tell the users that they can use the old syntax but that they would actually be using Galactic (implicitly). So we should basically alert the users that `sciml_train` still exists for legacy purposes but that they should be using Galactic instead?

ChrisRackauckas:
Yes, the plan is to fully deprecate `sciml_train` once the docs and tests are all updated.

mkg33:
Is this OK or shall I change anything?

mkg33:
Oh no, I must have missed the other tutorials that still use sciml_train. I'll also update them!

ChrisRackauckas:
I think all tutorials use it except like 2. So, it's gonna be a lot 😆

mkg33:
Apparently, I can't use the search function properly in my editor 🤦‍♂️ .

Issue #481, opened 2021-01-24
Do Multiple Shooting


Manas2030:
Training NN by splitting the datapoints and training on smaller intervals, to avoid being stuck in local minima while training.

Function I tried to train NN to predict values according to function, f(x) = x*sin(3*x)
Here is the code I used to get my observations:
```
using DiffEqFlux, OrdinaryDiffEq, Flux, Optim, Plots

u0 = Float32[0.0]
datasize = 50

tspan = (0.0f0,10.0f0)
tsteps = range(tspan[1],tspan[2],length = datasize)

ode_data = tsteps .* sin.( 3 .* tsteps)

dudt2 = FastChain((x, p) -> x.^3,
                  FastDense(1, 16, tanh),
                  FastDense(16, 1))
prob_neuralode = NeuralODE(dudt2, tspan, Vern7(), saveat = tsteps, abstol=1e-6, reltol=1e-6)

function predict_neuralode(p)
  Array(prob_neuralode(u0, p))
end

# x = number of groups we are dividing the dataset into
function loss_neuralode(p)
           pred = predict_neuralode(p)
           tot_loss = 0
           x = 5
           for i in 1:x
               loss = sum(abs, (ode_data[Int((i-1)*(datasize/x)+1):Int(i*(datasize/x))] .- pred[Int((i-1)*(datasize/x)+1):Int(i*(datasize/x))]))
               tot_loss += loss*loss
           end
           return tot_loss, pred
       end

iter = 0
callback = function (p, l, pred; doplot = true)
  global iter
  iter += 1

  display(l)
  if doplot
    # plot current prediction against data
    plt = scatter(tsteps[:], ode_data[:], label = "data")
    scatter!(plt, tsteps[:], pred[:], label = "prediction")
    display(plot(plt))
  end

  return false
end


result_neuralode = DiffEqFlux.sciml_train(loss_neuralode, prob_neuralode.p, ADAM(0.05), cb = callback, maxiters = 300)

callback(result_neuralode.minimizer,loss_neuralode(result_neuralode.minimizer)...;doplot=true)

function final_loss(p)
           return sum(abs2, (ode_data .- prob_neuralode(u0, p)))
end

print("Final Loss:",final_loss(result_neuralode.minimizer),"\n")
```
Observations:
Training on the whole dataset:
NN Architecture and hyperparameters used:
```
dudt2 = FastChain((x, p) -> x.^3,
                         FastDense(1, 16, tanh),
[...]

PR #505, opened 2021-03-11
Do multiple shooting

Manas2030:
This is in reference to issue #481

### DIFF
--- a/docs/src/examples/multiple_shooting.md
+++ b/docs/src/examples/multiple_shooting.md
@@ -1,10 +1,18 @@
 # Multiple Shooting
 
-In Multiple Shooting, the training data is split into overlapping intervals. The solver is then trained on individual intervals. If the end conditions of any interval co-incide with the initial conditions of the next immediate interval, then the joined/combined solution is same as solving on the whole dataset (without splitting).
+In Multiple Shooting, the training data is split into overlapping intervals.
+The solver is then trained on individual intervals. If the end conditions of any
+interval co-incide with the initial conditions of the next immediate interval,
+then the joined/combined solution is same as solving on the whole dataset
+(without splitting).
 
-To ensure that the overlapping part of two consecutive intervals co-incide, we add a penalizing term, `continuity_strength * absolute_value_of( prediction of last point of some group, i - prediction of first point of group i+1 )`, to the loss.
+To ensure that the overlapping part of two consecutive intervals co-incide,
+we add a penalizing term, `continuity_strength * absolute_value_of( prediction
+of last point of some group, i - prediction of first point of group i+1 )`, to
+the loss.
 
-Note that the `continuity_strength` should have a large positive value to add high penalities in case the solver predicts discontinuous values.
+Note that the `continuity_strength` should have a large positive value to add
+high penalities in case the solver predicts discontinuous values.
 
 
 The following is a working demo, using Multiple Shooting
@@ -68,7 +76,7 @@ callback = function (p, l, pred, predictions; doplot = true)
 end
 
 # Define parameters for Multiple Shooting
-grp_size_param = 5
+grp_size_param = 1
 loss_multiplier_param = 100
 
 neural_ode_f(u,p,t) = dudt2(u,p)
@@ -87,16 +95,24 @@ result_neuralode = DiffEqFlux.sciml_train(loss_neuralode, prob_neuralode.p,
                                           maxiters = 300)
 callback(result_neuralode.minimizer,loss_neuralode(result_neuralode.minimizer)...;doplot=true)
 
+result_neuralode_2 = DiffEqFlux.sciml_train(loss_neuralode, result_neuralode.minimizer,
+                                          BFGS(), cb = callback,
+                                          maxiters = 100, allow_f_increases=true)
+callback(result_neuralode_2.minimizer,loss_neuralode(result_neuralode_2.minimizer)...;doplot=true)
 
-``` 
+```
 Here's the plots that we get from above
 
-![pic](https://user-images.githubusercontent.com/58384989/111841264-b6488a00-8923-11eb-9648-3b4dfa73b833.PNG)
-The picture on the left shows how well our Neural Network does on a single shoot after training it through `multiple_shoot`.
-The picture on the right shows the predictions of each group (Notice that there are some overlapping points. These are the points we are trying to co-incide.)
+![pic](https://user-images.githubusercontent.com/58384989/111881194-6de9a480-89d5-11eb-8f21-6481d1e22521.PNG)
+The picture on the left shows how well our Neural Network does on a single shoot
+after training it through `multiple_shoot`.
+The picture on the right shows the predictions of each group (Notice that there
+are overlapping points as well. These are the points we are trying to co-incide.)
 
-Here is an output with `grp_size` = 30 (which is same as solving on the whole interval without splitting also called single shooting)
+Here is an output with `grp_size` = 30 (which is same as solving on the whole
+interval without splitting also called single shooting)
 
 ![pic_single_shoot3](https://user-images.githubusercontent.com/58384989/111843307-f0fff180-8926-11eb-9a06-2731113173bc.PNG)
 
-It is clear from the above picture, a single shoot doesn't perform very well with the ODE Problem we have and gets stuck in a local minima.
+It is clear from the above picture, a single shoot doesn't perform very well
+with the ODE Problem we have and gets stuck in a local minima.
--- a/docs/src/examples/multiple_shooting.md
+++ b/docs/src/examples/multiple_shooting.md
@@ -1,0 +1,102 @@
+# Multiple Shooting
+
+In Multiple Shooting, the training data is split into overlapping intervals. The solver is then trained on individual intervals. If the end conditions of any interval co-incide with the initial conditions of the next immediate interval, then the joined/combined solution is same as solving on the whole dataset (without splitting).
+
+To ensure that the overlapping part of two consecutive intervals co-incide, we add a penalizing term, `continuity_strength * absolute_value_of( prediction of last point of some group, i - prediction of first point of group i+1 )`, to the loss.
+
+Note that the `continuity_strength` should have a large positive value to add high penalities in case the solver predicts discontinuous values.
+
+
+The following is a working demo, using Multiple Shooting
+
+```julia
+using DiffEqFlux, OrdinaryDiffEq, Flux, Optim, Plots
+
+# Define initial conditions and timesteps
+datasize = 30
+u0 = Float32[2.0, 0.0]
+tspan = (0.0f0, 5.0f0)
+tsteps = range(tspan[1], tspan[2], length = datasize)
+
+
+# Get the data
+function trueODEfunc(du, u, p, t)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Needs to be `include`d at the top level, needs docs.

src/multiple_shooting.jl
```diff
@@ -0,0 +1,41 @@
...
+- `tsteps`: Timesteps on which ode_data was calculated.
+- `prob`: ODE problem that the Neural Network attempts to solve.
+- `loss_function`: Any arbitrary function to calculate loss.
+- `grp_size`: The group size achieved after splitting the ode_data into equal sizes.
+- `continuity_term`: Multiplying factor to ensure continuity of predictions throughout different groups.
+
+!!!note
+		The parameter 'continuity_term' should be a relatively big number to enforce a large penalty whenever the last point of any group doesn't coincide with the first point of next group.
+"""
+function loss_neuralode_helper(p::Array,ode_data::Array,tsteps::Array,prob::ODEProblem,loss_function::Function,grp_size::Integer=5,continuity_term::Integer=100)
```
Vaibhavdixit02, line 22: This should probably have "multiple shooting" in its name. Also, have you taken a look at https://github.com/SciML/DiffEqParamEstim.jl/blob/e99fb8233f48be99540dc9bf63ac226bb22c3c22/src/multiple_shooting_objective.jl#L24?

Manas2030: Hi, I'll change the name of the function to 'multiple_shooting'. And I took a look at the link now, do you have any suggestions from there? 😅

yewalenikhil65:
I am little new to these concepts. Can anyone please tell me how is `mini-batching` different from `multiple-shooting` ? I could gather that both involve chunks of datasets to be trained, but i couldn't see much difference.

ChrisRackauckas:
you don't solve the whole interval, and you do multiple batches in a loss call

yewalenikhil65:
> you don't solve the whole interval, and you do multiple batches in a loss call

ohh yes..So,
1) loss is summed up over all the batches in multiple-shooting, 
2) initial condition is also changed accordingly in multiple-shooting..unlike in mini-batching

ChrisRackauckas:
yes

src/multiple_shooting.jl
Vaibhavdixit02: Not really, just added it for your reference

src/multiple_shooting.jl
```diff
@@ -0,0 +1,41 @@
...
+
+		pred = solve(remake(prob, p = p, tspan = (tsteps[1],tsteps[datasize]), u0 = ode_data[:,1]), Tsit5(),saveat = tsteps)
+		return tot_loss, pred
+	end
+
+	for i in 1:grp_size-1:datasize-grp_size
+		pred = solve(remake(prob, p = p, tspan = (tsteps[i],tsteps[i+grp_size-1]), u0 = ode_data[:,i]), Tsit5(),saveat = tsteps[i:i+grp_size-1])
+		tot_loss += loss_function(ode_data[:,i:i+grp_size-1], pred[:,1:grp_size]) + (continuity_term * sum(abs,pred[:,grp_size] - ode_data[:,i+grp_size-1]))
+	end
+	pred = solve(remake(prob, p = p, tspan = (tsteps[1],tsteps[datasize]), u0 = ode_data[:,1]), Tsit5(),saveat = tsteps)
```
yewalenikhil65, line 39: Shouldn't the `Line 39`, be a concatenation of all the `pred` calculated inside the `for` loops, which is actually prediction corresponding to the `tot_loss` ?

yewalenikhil65 (commented):
Shouldn't the Line 39, be a concatenation of all the pred calculated inside the for loops, which is actually prediction corresponding to the tot_loss ?

Manas2030:
> Shouldn't the `Line 39`, be a concatenation of all the `pred` calculated inside the `for loops`, which is actually prediction corresponding to the `tot_loss` ?

Yes it indeed would be the actual predictions but I assume multiple shooting is just for training (i.e. calculating the loss) and the actual predictions (that is plotting the curve) would be done in a single shoot. Also the end points in any group overlap so there are 2 different predictions for them which would be weird to plot(?).

@ChrisRackauckas  What are your thoughts on this?

yewalenikhil65:
> > Shouldn't the `Line 39`, be a concatenation of all the `pred` calculated inside the `for loops`, which is actually prediction corresponding to the `tot_loss` ?
> 
Also the end points in any group overlap so there are 2 different predictions for them which would be weird to plot(?).
> 
1-2, 2-3..say these are the batches
1) wrong prediction at `2` while you train from `1-2`
2) but then this `prediction` at `2` will be rectified because you use the `ode_data` at `2` initial condition for `2-3`

I think, the in between points prediction would be  accurate, because that `endpoint` is your initial condition for the next prediction... only intermediate points within the intervals would be off the exact `ode_data`.. Correct ?

yewalenikhil65:
> Yes, I considered this and it's true we'll only have to overwrite end predictions of groups but while calculating loss we do sum up the offset of the point that we are overwriting. So it makes sense to display that offset as well if we are going to display it for the other remaining points.
> 
> Also, this will bring up discontinuties while plotting (which is fine but depends more on what the user would want)
> 
> And for the example you gave, it would plot the exact data we have, i.e. `ode_data`, if we overwrite each of the overlapping points (since every point is overlapping). So graphically our loss would always be 0 but this shouldn't happen

correct.. so i suggest this
**What seems then more correct to me is, we should concatenate all pred calculated inside for loop, but should drop the pred[,:1] for all i=2,3,4....n as they are just initial conditions.**

This won't bring discontinuities, and also it won't create `loss` to be zero prematurely.

Vaibhavdixit02:
I don't recall if there was a discussion around this before, you should go through the related commits/PRs @Manas2030 but it does look like were returning the merged solution here https://github.com/SciML/DiffEqParamEstim.jl/blob/e99fb8233f48be99540dc9bf63ac226bb22c3c22/src/multiple_shooting_objective.jl#L58-L63

Manas2030:
@yewalenikhil65  I wrote above comment reading your previous suggestion (before the edits) but for this one I think there would still be discontinuities. Here's a very naive example:

Suppose you are trying to predict the line, y = 100 at points x = 1 to 20 with grp_size/batchsize = 5
Suppose for first group your predictions lie above y = 100 with prediction at x = 1 being 100. So with some value of the parameters it's very much possible that for the next group predictions lie below y = 100 with prediction for x = 5 being 100.
This brings up a discontinuity.

Also there is nothing wrong with plotting discontinuities but I am not sure if it would be a better idea so honestly I was waiting for Chris's opinion on this 😅
Other than that it's a good suggestion, I don't see any or previous problems 😄 

@Vaibhavdixit02 I see. Do you think it would be a better idea if we add a parameter that let's the user decide what predictions they want? To be honest, it doesn't affect the training so anything should be fine and it's likely be used for only plotting purposes.

ChrisRackauckas:
Multiple shooting is likely just for prediction, though it would be good to have a way to visualize it, otherwise it can be hard to see what's going on with the training. Indeed, DiffEqParamEstim.jl's multiple shooting was good, but you couldn't do more than define a loss, get parameters, and question what went wrong if it didn't fit.

Multiple shooting is easy to visualize:

https://d3i71xaburhd42.cloudfront.net/4284951012483c6914a7b1bfd7930863cb47f5cf/5-Figure2-1.png

For the discontinuities, just have two points exactly at `t` and it'll handle the discontinuities. 

https://diffeq.sciml.ai/stable/features/callback_functions/#Example-1:-Interventions-at-Preset-Times

That has tons of plots with discontinuities. So I think giving someone a good way to visualize the multiple shooting fit would be pretty good.

#### Larger comment

The larger vision here is that DiffEqParamEstim.jl was our first attempt at a parameter estimation library, but the autodiff enhancements led to DiffEqFlux.jl which is a lot more general. Multiple shooting is one of the last reasons to use DiffEqParamEstim.jl, because that library was based around defining loss functions while DiffEqFlux is all about the user defining their own loss functions. Except, multiple shooting is a fancy loss function. How do we make it easy to use a fancy loss function while keeping the flexibility of DiffEqFlux, things like having control over which variables are used when checking against the data, and switching between forward and reverse mode approaches? That's the question to answer here.

I think this PR has done a generally good job, though it needs to remove some of those `::Array` annotations. But, somehow adding a visualization to multiple shooting so it can be animated similar to the tutorials would be great for teaching and for diagnosing fitting issues.

Manas2030:
> Needs to be `include`d at the top level, needs docs.

@ChrisRackauckas  What did you mean by top level? And for docs should I append it to https://diffeqflux.sciml.ai/dev/examples/local_minima/ or make a whole new tutorial? 😅

ChrisRackauckas:
I mean, this file isn't part of the library yet. It hasn't been `include`d into the module. Separately, it needs docs and tests.

Manas2030:
Oh okay, I see but where can I upload the code (as in which directory/location)? I am confused.

Vaibhavdixit02:
@Manas2030 what Chris is saying is that you need to have the `multiple_shooting.jl` file `include`-ed in the src/DiffEqFlux.jl file where we define the module, please take a look at that file and that should clear things up for you.

>And for docs should I append it to https://diffeqflux.sciml.ai/dev/examples/local_minima/ or make a whole new tutorial?

I think, it can go in there, but I would like to see a convincing example of where you actually benefit with multiple shooting so let's get to that from a different PR with some discussion.

This PR needs a test. (Take a look at other tests it should be very easy to add one)

Manas2030:
Ah, okay this makes it very clear. Thank you!

docs/src/examples/multiple_shooting.md
```diff
@@ -0,0 +1,102 @@
+# Multiple Shooting
+
+In Multiple Shooting, the training data is split into overlapping intervals. The solver is then trained on individual intervals. If the end conditions of any interval co-incide with the initial conditions of the next immediate interval, then the joined/combined solution is same as solving on the whole dataset (without splitting).
```
ChrisRackauckas, line 3: Line lengths

docs/src/examples/multiple_shooting.md
```diff
@@ -0,0 +1,102 @@
...
+function loss_function_param(ode_data, pred):: Float32
+	return sum(abs2, (ode_data .- pred))^2
+end
+
+function loss_neuralode(p)
+	return multiple_shoot(p, ode_data, tsteps, prob_param, loss_function_param, grp_size_param, loss_multiplier_param)
+end
+
+result_neuralode = DiffEqFlux.sciml_train(loss_neuralode, prob_neuralode.p,
+                                          ADAM(0.05), cb = callback,
```
ChrisRackauckas, line 86: you're missing the BFGS that comes after ADAM

docs/src/examples/multiple_shooting.md
Manas2030: Do you mean that I should split this into different lines? (and also preferably write shorter lines in future)

ChrisRackauckas: Your text editor should show you the standard maximum line length. Break lines at that spot.

PR #519, opened 2021-04-02
Optimize the Neural SDE Demo

avik-pal:
Batching the generation of multiple trajectories improves the forward pass (for n = 100) from 160ms to 11ms. On GPUs where repeated repeats would be expensive, we could initialize the batched u once.


ChrisRackauckas (approved):

ChrisRackauckas:
Great idea.

PR #521, opened 2021-04-09
Use other solvers than Tsit5 in multiple_shoot 

adrhill:
Previously, Tsit5 was hardcoded in multiple_shooting.jl. The function now takes a solver `solver::DiffEqBase.AbstractODEAlgorithm`, which is called on the solve.

Other changes
* fixed parameter name in docstring
* fixed assert message and replaced it with DomainError
* relaxed types
* removed `grp_size` default value as it could lead to domain errors
* added test for DomainErrors


ChrisRackauckas (approved):
awesome.

PR #523, opened 2021-04-10
Upper bound adapt.jl

ChrisRackauckas:


maleadt:
Instead of directly upper bounding, why not file an issue with an MWE and/or attempt to fix Adapt? I'm happy to tag a hotfix release ASAP. This will now only hold back packages that depend on Adapt 3.3, like CUDA 3.0.

ChrisRackauckas:
This isn't meant to merge. It was just a test to see if everything in SciML gets fixed by forcing Adapt 3.2 (since DiffEqFlux is a pretty big stress test). That indeed turns out to be the case. I am quite confused how Adapt 3.3 does anything remotely bad though: it looks so benign that I am quite confused. But this is good evidence that it's the problem.

ChrisRackauckas:
Merged an upper bound to get a separate PR merged, but am hoping to pull that before tagging.

Issue #522, opened 2021-04-09
Possible Indexing error in multiple_shoot


adrhill:
When calling `multiple_shoot` with `grp_size=2`, the following line is called:
https://github.com/SciML/DiffEqFlux.jl/blob/ee4937f33148dfeb7b71914763b9b8b86c4dc094/src/multiple_shooting.jl#L50

Which looks equal to calling the function with `grp_size=1`:
https://github.com/SciML/DiffEqFlux.jl/blob/ee4937f33148dfeb7b71914763b9b8b86c4dc094/src/multiple_shooting.jl#L35-L37

I also feel like the three cases (`grp_size==datasize`, `grp_size==1`, else) are not really necessary, as they all perform the same calculation. 

___ 

On another note, `test/multiple_shoot.jl` doesn't seem to be included in `runtests.jl` yet.

ChrisRackauckas:
Yes, those cases probably aren't necessary.

yewalenikhil65:
Copy pasting the code now also seems to throw error on Tsit5 not defined ?
I am sure it was working fine a couple of weeks back though
![image](https://user-images.githubusercontent.com/26343789/114272154-97c14480-9a32-11eb-8fcb-afd75c7354bf.png)

adrhill:
I think PR #521 fixes this.

yewalenikhil65:
> I think PR #521 fixes this.

Awesome!

adrhill:
This was closed by #535.

PR #535, opened 2021-04-22
Fix & simplify multiple shooting (#522)

adrhill:
## Fix indexing
As a mentioned in #522, the indexing happening in the comprehensions of `multiple_shoot` seemed error-prone. Additionally, the cases weren't necessary.

I've simplified the code by introducing a helper function `group_ranges` with matching tests. 
It splits a dataset of size `datasize` (e.g. 10)  into ranges of length `group_size` (e.g. 5):
```julia-repl
julia> group_ranges(10, 5)
3-element Vector{UnitRange{Int64}}:
 1:5
 5:9
 9:10
```
With these ranges, the predictions can be calculated in a comprehension as it was previously done.
As you can see in the last range, it now also makes sure that **no data gets discarded** in the last group, which previously wasn't the case.

##  Remove single shooting
I've also chosen to remove the single shooting prediction that was previously run at the end:
https://github.com/SciML/DiffEqFlux.jl/blob/ee4937f33148dfeb7b71914763b9b8b86c4dc094/src/multiple_shooting.jl#L58-L60

1. It slows down training
2. One use-case of multiple shooting is to avoid unstable dynamics by integrating over shorter timespans. This re-introduced that instability.

## Minimum group size
I've also chosen to set the minimum `group_size` to `>=2`. As far as I understand, using a size of 1, the solution is simply given as the initial condition`u0` in every group and nothing is learned.

### DIFF
--- a/docs/src/examples/multiple_shooting.md
+++ b/docs/src/examples/multiple_shooting.md
@@ -78,7 +78,7 @@ group_size = 3
 continuity_term = 200
 
 function loss_function(data, pred)
-	return sum(abs2, data - pred)^2
+	return sum(abs2, data - pred)
 end
 
 function loss_multiple_shooting(p)
--- a/docs/src/examples/multiple_shooting.md
+++ b/docs/src/examples/multiple_shooting.md
@@ -7,11 +7,11 @@ then the joined/combined solution is same as solving on the whole dataset
 (without splitting).
 
 To ensure that the overlapping part of two consecutive intervals coincide,
-we add a penalizing term, `continuity_term * absolute_value_of( prediction
-of last point of some group, i - prediction of first point of group i+1 )`, to
+we add a penalizing term, `continuity_term * absolute_value_of(prediction
+of last point of group i - prediction of first point of group i+1)`, to
 the loss.
 
-Note that the `continuity_strength` should have a large positive value to add
+Note that the `continuity_term` should have a large positive value to add
 high penalties in case the solver predicts discontinuous values.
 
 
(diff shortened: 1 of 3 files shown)


adrhill:
Here's some eye-candy from the example in the docs: 

(from the ADAM part of the training)

ChrisRackauckas:
beautiful

docs/src/examples/multiple_shooting.md
```diff
@@ -37,68 +38,62 @@ ode_data = Array(solve(prob_trueode, Tsit5(), saveat = tsteps))
...
 
-callback = function (p, l, pred, predictions; doplot = true)
+# Animate training
+anim = Animation()
+callback = function (p, l, preds; doplot = true)
   display(l)
   if doplot
 	# plot the original data
-	plt = scatter(tsteps[1:size(pred,2)], ode_data[1,1:size(pred,2)], label = "data")
+	plt = scatter(tsteps[:], ode_data[1,:], label = "Data")
```
ChrisRackauckas, line 65:
```suggestion
	plt = scatter(tsteps, ode_data[1,:], label = "Data")
```

docs/src/examples/multiple_shooting.md
```diff
@@ -37,68 +38,62 @@ ode_data = Array(solve(prob_trueode, Tsit5(), saveat = tsteps))
...
-
-neural_ode_f(u,p,t) = dudt2(u,p)
-prob_param = ODEProblem(neural_ode_f, u0, tspan, initial_params(dudt2))
+group_size = 3
+continuity_term = 200
 
-function loss_function_param(ode_data, pred):: Float32
-	return sum(abs2, (ode_data .- pred))^2
+function loss_function(data, pred)
+	return sum(abs2, data - pred)^2
```
ChrisRackauckas, line 81:
```suggestion
	return sum(abs2, data - pred)
```

ChrisRackauckas (approved):
This is a pretty major improvement to the API. Thanks!

ChrisRackauckas:
>from the updated example in the docs:

Link it into the docs?

adrhill:
Sure, do you want to link to Imgur or do you want the gif to be saved into some asset folder?
I wasn't sure whether the examples in the docs are run or not, so I uncommented the last line that saves the gif.

ChrisRackauckas:
Just use the address from this PR thread: ``

adrhill:
I think Codecov is complaining because `src/multiple_shooting.jl` still isn't included in `runtests`. I'm not sure which group of tests I should add it to.

ChrisRackauckas:
Yeah Codecov is kind of broken 🤷

adrhill:
I've made some additional changes to allow custom "`continuity_loss`" functions via multiple dispatch. Should I commit them to this PR?

ChrisRackauckas:
Do it in the next.

Issue #509, opened 2021-03-13
Liquid time-constant networks


ChrisRackauckas:
https://arxiv.org/abs/2006.04439
https://github.com/raminmh/liquid_time_constant_networks

PR #536, opened 2021-04-22
adds ltc layer and its example

manyfeatures:
PR adds Liquid Constant Time Layer [#509](https://github.com/SciML/DiffEqFlux.jl/issues/509#issue-830860839).

Original [repo1](https://github.com/raminmh/liquid_time_constant_networks) and [repo2](https://github.com/mlech26l/keras-ncp) contain the code which is consistent with biological inspired NCP architecture.   

Screenshot from 2021-04-22 18-53-17

The current implementation differs from them and corresponds to the plain description from the paper
Screenshot from 2021-04-22 18-59-21
 
To ensure the stability the output of the network is clipped to be in range between -1 and 1

### DIFF
--- a/docs/src/examples/LTC_layer.md
+++ b/docs/src/examples/LTC_layer.md
@@ -1,0 +1,50 @@
+using DiffEqFlux, Flux, Plots, Statistics
+using Random
+Random.seed!(1234); # Fix seed
+
+N = 48
+π_32 = Float32(π)
+t = range(0.0f0,stop=3π_32, length = N)
+sin_t = sin.(t)
+cos_t = cos.(t)
+data_x = vcat(reshape(sin_t,(1,N)), reshape(cos_t,(1,N)))
+data_y = reshape(sin.(range(0.0f0,stop=6π_32, length = N)), (1, N))
+data_x = [data_x[:,i] for i=1:N]
+data_y = [[data_y[i]] for i=1:N]
+
+println(size(data_x))
+println(size(data_y))
+
+
+m = Chain(LTC(2,32), Dense(32,1,x->x))
+function loss_(x,y)
+    diff = (m.(x) .- y)
+    diff = [diff[i][1] for i=1:N]
+    mean(abs2.(diff))
+end
+
+#callback function to observe training
+cb = function ()
+  cur_pred = m.(data_x)
+  pl = plot(t, [data_y[i][1] for i=1:length(data_y)], label="data")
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Looks great but I won't be able to review until tomorrow

Vaibhavdixit02:
Couple of points, I am still going through it and will get back with more informed feedback later.

1. I can see that in the python repo they are doing the evolution of the ode with the loop. I am wondering if we can instead use the `solve` for it by defining the diff eq mentioned in eq 1 directly. My interpretation is the `LTC` should look similar to the `NeuralODE` and be something like

```
struct LTC{M,P,RE,T,TA,AB,A,K} <: NeuralDELayer
    model::M
    p::P
    re::RE
    tspan::T
    τ::TA
    A::AB
    args::A
    kwargs::K
```
where user gives you the NN as the `model` and we internally create the ODEProblem with is as (`n` is the `LTC` layer below)

```
dudt_(u,p,t) = -(1/n.τ + n.re(p)(u)) * u +  n.re(p)(u)*n.A
ff = ODEFunction{false}(dudt_,tgrad=basic_tgrad)
prob = ODEProblem{false}(ff,x,getfield(n,:tspan),p)
```

2. The current example you added doesn't seem to have any benefit of using LTC and instead actually gives much higher loss for the same number of epochs (400 that you have used) as compared to just a dense layer.

```
##LTC
epoch = 400
loss_(data_x, data_y) = 0.0020240898579002888

## changing m to m = Chain(Dense(2,32, tanh), Dense(32,1,x->x)) in the example
epoch = 400
loss_(data_x, data_y) = 1.0341962f-5
```
 
I think we can use this as the test, but for an example we would want to recreate of the experiments from the paper, probably MNIST since we have some existing code for it that should help it get up and running.

ChrisRackauckas:
Agree 100% on @Vaibhavdixit02 's points.

Vaibhavdixit02:
@manyfeatures just checking if this is something you are still working on?

manyfeatures:
@Vaibhavdixit02 Yep, I'll try to remaster the PR in a couple of days

Vaibhavdixit02:
Great 👍

manyfeatures:
I've got some problem because can't limit the hidden state amplitude if I transform the problem in NeuralODE layer. I'll describe the problem on discourse

manyfeatures:
I created the [post](https://discourse.julialang.org/t/custom-rnn-cell-help-needed-to-improve-performance/60536) with code and also will examine it lately as well

PR #546, opened 2021-04-30
Add solver kwargs to multiple shooting

adrhill:
Allows setting `sensalg` and tolerances  when using `multiple_shoot`.


src/multiple_shooting.jl
```diff
@@ -91,7 +100,7 @@ function multiple_shoot(
     loss_function::Function,
     solver::DiffEqBase.AbstractODEAlgorithm,
     group_size::Integer;
-    continuity_term::Real=100,
```
ChrisRackauckas, line 94: why delete?

PR #559, opened 2021-05-20
Revert "Update divergence.md"

ChrisRackauckas:
Reverts SciML/DiffEqFlux.jl#549


ChrisRackauckas:
@Vaibhavdixit02 reverting for now until we get something in the docs that works.

PR #574, opened 2021-06-20
Simplify and improve sciml_train heuristics

ChrisRackauckas:
DiffEqFlux.jl is a high level user-facing interface for scientific machine learning, inverse problems, and parameter estimation. At one point it played a much larger role in defining optimizer interfaces, adjoint methods, and more, but now all of that has bee shipped off to dedicated packages.

With that said, what is left of DiffEqFlux.jl is that it is the high level interface. It has nice tutorials, it has helper functions, and it has heuristics. Because that's it's role, it should go all in for doing that right.

This PR does just that. DiffEqFlux is the DifferentialEquations of parameter estimation, mixed with the high level layer tooling of something like torchdyn. Quick summary of the changes to make that smooth:

1. sciml_train is given heuristics to choose multi-optimizer chains and switch between AD modes automatically.
2. More reexports are done so that the `using` chain is now very simple. 
3. Makes sure that the underlying packages (GalacticOptim.jl, DiffEqSensitivity.jl, etc.) are highlighted for users to go deeper.
4. Updates the tutorials to streamline them.

The result is that the user code is a lot simpler, and a lot of it is a lot more optimized due to improved heuristics.


ChrisRackauckas:
node

PR #581, opened 2021-06-23
Use return codes in multiple shooting

adrhill:
Previously, failure of integration of one of the groups in multiple shooting (e.g. due to `:DtLessThanMin`) lead to an interruption of training as there would be a dimension mismatch between the training data and the prediction.
 
## Changes
1. In this PR,  infinite loss is returned on failure of integration by using the solution return codes. 
    I wasn't sure how to add tests for this, so let me know if you have an idea.

1. I also included `test/multiple_shoot.jl` in `test/runtest.jl`, as the tests previously weren't run in CI. I wasn't sure which section to include them in, so you might want to change this.

### DIFF
--- a/src/multiple_shooting.jl
+++ b/src/multiple_shooting.jl
@@ -58,21 +58,26 @@ function multiple_shoot(
     ranges = group_ranges(datasize, group_size)
 
     # Multiple shooting predictions
-    group_predictions = [
-        Array(
-            solve(
-                remake(
-                    prob;
-                    p=p,
-                    tspan=(tsteps[first(rg)], tsteps[last(rg)]),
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):
yes good idea.

ChrisRackauckas:
I'm not sure how to test it either. Maybe do LV with some weirdly bad starting parameters?

adrhill:
I added a test setting the solver options to `maxiters=1`, which seems to do the job.
Is there any way to suppress the expected/intended warning? Or should I just leave it like this?

ChrisRackauckas:
The user can pass the `verbose=false` kwarg to remove it.

PR #590, opened 2021-07-19
Tutorial edits

Gregliest:
This PR includes fixes and edits to the tutorials. I also reworded a few things, and added clarification in a couple of places. I added a bunch of comments to the code in the  `NeuralODE.sciml_train` tutorial, with the thinking that this is probably people's first tutorial for NeuralODE's, and so it should probably be as detailed as possible. 

Feel free to change or ignore as you see fit.


ChrisRackauckas:
Thanks, great edits!

PR #592, opened 2021-07-20
use Flux.zeros32

DhairyaLGandhi:
Fixes

```julia
julia> FastDense(1,12).initial_params()
13-element Vector{Union{Nothing, Float32}}:
  0.6627236f0
  0.676602f0
 -0.59016824f0
 -0.53156716f0
 -0.58506316f0
  0.00657402f0
  ⋮
 -0.58002657f0
 -0.092621975f0
  0.30483305f0
  0.30343133f0
   nothing
```

This still needs a fix in Flux.

### DIFF
--- a/src/fast_layers.jl
+++ b/src/fast_layers.jl
@@ -41,7 +41,7 @@ struct FastDense{F,F2} <: FastLayer
   initial_params::F2
   bias::Bool
   function FastDense(in::Integer, out::Integer, σ = identity;
-                 bias = true, initW = Flux.glorot_uniform, initb = Flux.zeros3232)
+                 bias = true, initW = Flux.glorot_uniform, initb = Flux.zeros32)
     temp = ((bias == false) ? vcat(vec(initW(out, in))) : vcat(vec(initW(out, in)),initb(out)))
     initial_params() = temp
     new{typeof(σ),typeof(initial_params)}(out,in,σ,initial_params,bias)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
@DhairyaLGandhi I tracked it down to:

```julia
using DiffEqFlux, Flux, OrdinaryDiffEq, DiffEqSensitivity
n = 1  # number of ODEs
x = rand(n, 5)
y = rand(n, 5)

tspan = (0.0, 1.0)
NN = Dense(n, n, tanh)
nODE = NeuralODE(NN, tspan, ROCK4(), reltol=1e-4, saveat=[tspan[end]], sensealg=InterpolatingAdjoint())
function loss_function(θ)
    z = nODE(x, θ)
    Flux.mse(y, z)
end
size(nODE(x, nODE.p)[:,1]) == size(y)
yy,back = Zygote.pullback(loss_function,nODE.p)
back(1)
```

I thought I fixed it with:

https://github.com/SciML/RecursiveArrayTools.jl/pull/164

For some reason, the new getindex fallbacks were causing things to fail. But now, this overload catches it but has an issue. What was the old dispatch it was hitting?

DhairyaLGandhi:
https://github.com/FluxML/Zygote.jl/blob/975923915c6533abd278902aec6dfe07f1929f46/src/lib/lib.jl#L103

Is this the one you're thinking of?

ChrisRackauckas:
No, the getindex methods on vector must've recently changed.

DhairyaLGandhi:
Should we be pulling out the last index here? Seems like something that shouldn't need manual handling. ref https://github.com/FluxML/Flux.jl/pull/1636

Issue #601, opened 2021-08-09
`multiple_shoot` for `EnsembleProblem`


vboussange:
Hey guys, 
I need to do multiple shooting with an `EnsembleProblem`.
Does it make sense for you to overload `multiple_shoot` for `EnsembleProblem`? 
I am happy to propose a PR 😊

Vaibhavdixit02:
I think it would be useful but not sure what that might look like, since it could probably work in a couple of different ways, do you have any use case for it? Feel free to open the PR and we can discuss it better there

vboussange:
Sure, I will open a PR any time soon

PR #602, opened 2021-08-10
`multiple_shoot` for `EnsembleProblem`

vboussange:
Hi,
This PR overloads the `multiple_shoot` function, so that it can be called with problems of type `EnsembleProblem`.
This is quite useful 😀 
It is a suggestion for issue #601.


src/multiple_shooting.jl
```diff
@@ -11,27 +11,47 @@ multiple_shoot(p, ode_data, tsteps, prob, loss_function, solver, group_size;
...
+    - `tsteps`: Timesteps on which ode_data was calculated.
+    - `prob`: ODE problem that the Neural Network attempts to solve.
+    - `loss_function`: Any arbitrary function to calculate loss.
+    - `continuity_loss`: Function that takes states ``\\hat{u}_{end}`` of group ``k`` and
+    ``u_{0}`` of group ``k+1`` as input and calculates prediction continuity loss
+    between them.
+    If no custom `continuity_loss` is specified, `sum(abs, û_end - u_0)` is used.
+    - `solver`: ODE Solver algorithm.
+    - `group_size`: The group size achieved after splitting the ode_data into equal sizes.
+    - `continuity_term`: Weight term to ensure continuity of predictions throughout
```
ChrisRackauckas, line 34: put this in a separate docstring on the other dispatch.

src/multiple_shooting.jl
```diff
@@ -126,6 +146,118 @@ function multiple_shoot(
...
 end
 
+function multiple_shoot(
+    p::AbstractArray,
+    ode_data::AbstractArray,
+    tsteps::AbstractArray,
+    ensembleprob::EnsembleProblem,
+    ensemblealg::SciMLBase.BasicEnsembleAlgorithm,
+    loss_function::Function,
+    continuity_loss::Function,
```
ChrisRackauckas, line 156: This stops specialization BTW

src/multiple_shooting.jl
```diff
@@ -126,6 +146,118 @@ function multiple_shoot(
...
+        loss_function,
+        continuity_loss,
+        solver,
+        group_size;
+        continuity_term,
+        kwargs...
+    )
+    
+end
+
```
ChrisRackauckas, line 260: we should really move the optional argument to last

src/multiple_shooting.jl
```diff
@@ -126,6 +146,118 @@ function multiple_shoot(
...
+    solver::DiffEqBase.AbstractODEAlgorithm,
+    group_size::Integer;
+    continuity_term::Real=100,
+    kwargs...
+)
+    # Continuity loss between last state in previous prediction
+    # and current initial condition in ode_data
+    function continuity_loss(û_end, u_0)
+        return sum(abs, û_end - u_0)
+    end
```
ChrisRackauckas, line 243: Define this as a generic function so it precompiles.

src/multiple_shooting.jl
vboussange: I have removed  the type specification `::Function`. Is that what you had in mind?

src/multiple_shooting.jl
vboussange: I am not sure what you mean by that.

ChrisRackauckas: the reason for this dispatch is to handle the optional argument, but it really shouldn't be one of the ones in the middle then.

PR #620, opened 2021-09-17
code cleaning for ffjord

prbzrg:
I used outer constructor and removed `reshape`, also removed `return`. `cat` in dim-1 is `vcat` so I replaced it


ChrisRackauckas:
Thanks!

Issue #615, opened 2021-09-02
ERROR: LoadError: MethodError: no method matching Float32(::ReverseDiff.TrackedReal{Float64, Float32, Nothing})


prbzrg:
When I run this code:
```julia
using DiffEqFlux, DifferentialEquations, GalacticOptim, Distributions

nn = Chain(
    Dense(1, 3, tanh),
    Dense(3, 1, tanh),
) |> f32
tspan = (0.0f0, 10.0f0)
ffjord_mdl = FFJORD(nn, tspan, Tsit5())

data_dist = Normal(6.0f0, 0.7f0)
train_data = rand(data_dist, 1, 100)

function loss(θ)
    logpx, λ₁, λ₂ = ffjord_mdl(train_data, θ)
    -mean(logpx)
end

adtype = GalacticOptim.AutoReverseDiff()
res1 = DiffEqFlux.sciml_train(loss, ffjord_mdl.p, ADAM(0.1), adtype; maxiters=100)
```
I get this error:
```julia
ERROR: LoadError: MethodError: no method matching Float32(::ReverseDiff.TrackedReal{Float64, Float32, Nothing})
Closest candidates are:
  (::Type{T})(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:200
  (::Type{T})(::T) where T<:Number at boot.jl:760
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at char.jl:50
  ...
Stacktrace:
  [1] _broadcast_getindex_evalf
    @ .\broadcast.jl:648 [inlined]
  [2] _broadcast_getindex
    @ .\broadcast.jl:621 [inlined]
  [3] getindex
    @ .\broadcast.jl:575 [inlined]
  [4] copy
    @ .\broadcast.jl:922 [inlined]
  [5] materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, Type{Float32}, Tuple{Matrix{ReverseDiff.TrackedReal{Float64, Float32, Nothing}}}})
    @ Base.Broadcast .\broadcast.jl:883
  [6] WARNING: both Flux and Iterators export "flatten"; uses of it in module DiffEqFlux must be qualified
WARNING: both Flux and Distributions export "params"; uses of it in module DiffEqFlux must be qualified
[...]

Issue #610, opened 2021-08-24
ERROR: LoadError: Cannot determine ordering of Dual tags ForwardDiff.Tag


prbzrg:
When I run this code:
```julia
using DiffEqFlux, DifferentialEquations, GalacticOptim, Distributions

nn = Chain(
    Dense(1, 3, tanh),
    Dense(3, 1, tanh),
) |> f32
tspan = (0.0f0, 10.0f0)
ffjord_mdl = FFJORD(nn, tspan, Tsit5())

data_dist = Normal(6.0f0, 0.7f0)
train_data = rand(data_dist, 1, 100)

function loss(θ)
    logpx, λ₁, λ₂ = ffjord_mdl(train_data, θ)
    -mean(logpx)
end

adtype = GalacticOptim.AutoForwardDiff()
res1 = DiffEqFlux.sciml_train(loss, ffjord_mdl.p, ADAM(0.1), adtype; maxiters=100)
```
I get this error:
```julia
ERROR: LoadError: WARNING: both Flux and Iterators export "flatten"; uses of it in module DiffEqFlux must be qualified
WARNING: both Flux and Distributions export "params"; uses of it in module DiffEqFlux must be qualified
Cannot determine ordering of Dual tags ForwardDiff.Tag{GalacticOptim.var"#117#133"{Tuple{}, GalacticOptim.var"#115#131"{OptimizationFunction{true, GalacticOptim.AutoForwardDiff{nothing}, DiffEqFlux.var"#74#79"{typeof(loss)}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing}, Nothing}}, Float32} and Nothing
Stacktrace:
  [1] ≺(a::Type, b::Type)
    @ ForwardDiff C:\Users\Hossein Pourbozorg\.julia\packages\ForwardDiff\5gUap\src\dual.jl:49
  [2] -
    @ C:\Users\Hossein Pourbozorg\.julia\packages\ForwardDiff\5gUap\src\dual.jl:135 [inlined]
  [3] _broadcast_getindex_evalf
    @ .\broadcast.jl:648 [inlined]
  [4] _broadcast_getindex
    @ .\broadcast.jl:621 [inlined]
  [5] getindex
    @ .\broadcast.jl:575 [inlined]
  [6] copy
    @ .\broadcast.jl:922 [inlined]
[...]

Issue #588, opened 2021-07-14
ERROR: UndefVarError: eJ not defined 


prbzrg:
When I run this code:
```julia
using DiffEqFlux, DifferentialEquations, Distributions, GalacticOptim, CUDA

ffjord_test = FFJORD(Chain(
    Dense(1, 4, tanh),
    Dense(4, 1, tanh),
), (0.0, 10.0))
train_data = rand(Normal(7.5, 0.75), 1, 100)

loss(θ) = -mean(first(ffjord_test(train_data, θ)))

res = DiffEqFlux.sciml_train(loss, ffjord_test.p, maxiters=100)

ffjord_test(train_data, regularize=true, monte_carlo=false)
```
I get this error:
```julia
WARNING: Method definition num_threads() in module VectorizationBase at C:\Users\Hossein Pourbozorg\.julia\packages\VectorizationBase\iB3H4\src\topology.jl:42 overwritten at C:\Users\Hossein Pourbozorg\.julia\packages\VectorizationBase\iB3H4\src\topology.jl:95.
WARNING: Method definition log(Float64) in module Math at special/log.jl:253 overwritten in module CUDA at C:\Users\Hossein Pourbozorg\.julia\packages\CUDA\02Kjq\src\device\intrinsics\math.jl:77.
WARNING: Method definition log(Float32) in module Math at special/log.jl:289 overwritten in module CUDA at C:\Users\Hossein Pourbozorg\.julia\packages\CUDA\02Kjq\src\device\intrinsics\math.jl:78.
WARNING: Method definition log_fast(Float32) in module FastMath at fastmath.jl:295 overwritten in module CUDA at C:\Users\Hossein Pourbozorg\.julia\packages\CUDA\02Kjq\src\device\intrinsics\math.jl:79.
WARNING: Method definition log10(Float64) in module Math at math.jl:575 overwritten in module CUDA at C:\Users\Hossein Pourbozorg\.julia\packages\CUDA\02Kjq\src\device\intrinsics\math.jl:81.
WARNING: Method definition log10(Float32) in module Math at math.jl:576 overwritten in module CUDA at C:\Users\Hossein Pourbozorg\.julia\packages\CUDA\02Kjq\src\device\intrinsics\math.jl:82.
[...]

PR #621, opened 2021-09-17
add smoke test for ffjord

prbzrg:
I just added some smoke tests for ffjord to reveal errors related to some adtypes
It reveals errors that reported in #588, #610, #615, #624
This PR can be merged when these errors fixed

### DIFF
--- a/test/cnf_test.jl
+++ b/test/cnf_test.jl
@@ -18,6 +18,30 @@ function cb(p, l)
     false
 end
 
+# smoke test
+@testset "Smoke test for adtype=$adtype & regularize=$rglrz & monte_carlo=$mnt_crl" for
+        adtype in [GalacticOptim.AutoForwardDiff(), GalacticOptim.AutoReverseDiff(), GalacticOptim.AutoTracker(),
+                   GalacticOptim.AutoZygote(), GalacticOptim.AutoFiniteDiff()],
+        rglrz in [false, true],
+        mnt_crl in [false, true]
+    nn = Chain(
+        Dense(1, 1, tanh),
+    ) |> f32
+    tspan = (0.0f0, 1.0f0)
+    ffjord_mdl = FFJORD(nn, tspan, Tsit5())
+
+    data_dist = Beta(2.0f0, 2.0f0)
+    train_data = rand(data_dist, 1, 100)
+
+    function loss(θ)
+        logpx, λ₁, λ₂ = ffjord_mdl(train_data, θ; regularize=rglrz, monte_carlo=mnt_crl)
+        -mean(logpx)
+    end
+
+    res1 = DiffEqFlux.sciml_train(loss, ffjord_mdl.p, ADAM(0.1), adtype; maxiters=10)
+    @test true
+end
+
 ###
 # test for default base distribution and deterministic trace CNF
 ###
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Instead, just set `@test_broken` for now.

prbzrg:
Should I expand the tests and unroll the loop?

ChrisRackauckas:
Are all of the tests broken?

prbzrg:
Regardless of `adtype`, using
```julia
regularize=true & monte_carlo=false
```
make errors and using
```julia
GalacticOptim.AutoForwardDiff()
GalacticOptim.AutoReverseDiff()
GalacticOptim.AutoTracker()
```
as `adtype` is also broken.
Overall, 14 of 20 tests are broken.

ChrisRackauckas:
>as adtype is also broken.

What do you mean?

prbzrg:
In each #610, #615, #624, I tested the same code with different `adtype` and reported errors.
`AutoZygote` and `AutoFiniteDiff` works fine, but other `adtype`s  make different errors.

ChrisRackauckas:
oh with FFJORD, yes.

ChrisRackauckas:
It looks like this still isn't catching some of the errors?

prbzrg:
If we only use `@test`, it could reproduce all four issues, but also make testing fail, you can see the errors in CI of first commit:
https://github.com/SciML/DiffEqFlux.jl/runs/3628543892?check_suite_focus=true#step:6:312
https://github.com/SciML/DiffEqFlux.jl/runs/3628543892?check_suite_focus=true#step:6:636
https://github.com/SciML/DiffEqFlux.jl/runs/3628543892?check_suite_focus=true#step:6:969
https://github.com/SciML/DiffEqFlux.jl/runs/3628543892?check_suite_focus=true#step:6:1350

I think we have two choices:

1. Only use `@test` and wait until all four issues fixed to merge this.
2. Use both `@test` and `@test_broken` by unrolling the loop. We can merge this sooner, but it doesn't log errors to CI, and it increases line of codes (more code to read and review)

I don't know whether there is any other choices. By second one, if someone wants to fix the errors, just changes it to `@test` and investigate the problems, but I prefer shorter codes.
Which way should we go?

ChrisRackauckas:
(2) is a better way to go, because it's always better to fix and change from broken than have things documented in unmerged PRs.

prbzrg:
OK, I made the loop unrolled and also converted section comments to `testset`s.
Furthermore, I notice something; there is a test for FFJORD with regularizers
https://github.com/SciML/DiffEqFlux.jl/blob/34c022fe0a829ad5a54499d8aa97355ecadb6e36/test/cnf_test.jl#L103-L127
but it doesn't use `regularize=true`!

prbzrg:
I want to update other tests to more look like tests of this PR:

1. Use `Float32` distribution instead of converting samples 
2. Use `DiffEqFlux.sciml_train` instead of `OptimizationFunction -> OptimizationProblem -> solve`

Can I update other tests in this PR?

prbzrg:
I made the updates. If you disagree, I will move them to new PR.

prbzrg:
Tests are failing. I just undo two last commits and will change other tests in a new PR.

ChrisRackauckas:
Thanks!

PR #627, opened 2021-09-24
add FFJORD sampler

prbzrg:
I created a Distribution based of `FFJORD`, so we can have sampling and pdf.
I noticed a mistake that didn't have test: constructing FFJORD with learned parameters.
Likewise, I added tests for sampling and pdf that also test constructing FFJORD with learned parameters.


ChrisRackauckas:
Oh cool idea! Actually, while you're add it, can you add a doc manual page on FFJORD? It has a tutorial but no manual, and we probably need to add this new Distribution to that as well.

prbzrg:
OK, I will do it.
IMO, let's discuss the manual page in a new issue and only focus on FFJORDDistribution in this PR.

ChrisRackauckas:
thanks!

PR #630, opened 2021-09-27
Move `sensealg` to functions arguments

prbzrg:
Now, we are able to use a different `sensealg`. However, we can remove `sensealg` to use default value and anyone can change it in `kwargs`, but I guess specified `sensealg`s have better performance.

### DIFF
--- a/src/ffjord.jl
+++ b/src/ffjord.jl
@@ -71,10 +71,9 @@ function cnf(du, u, p, t, re)
     du[end] = -trace_jac
 end
 
-function (n::DeterministicCNF)(x, p=n.p)
+function (n::DeterministicCNF)(x, p=n.p; sensealg=InterpolatingAdjoint(autojacvec=false))
     cnf_(du, u, p, t) = cnf(du, u, p, t, n.re)
     prob = ODEProblem{true}(cnf_, vcat(x, zero(eltype(x))), n.tspan, p)
-    sensealg = InterpolatingAdjoint(autojacvec=false)
     pred = solve(prob, n.args...; sensealg, n.kwargs...)[:, end]
     pz = n.basedist
     z = pred[1:end - 1]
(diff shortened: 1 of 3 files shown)


src/ffjord.jl
```diff
@@ -71,10 +71,9 @@ function cnf(du, u, p, t, re)
     du[end] = -trace_jac
 end
 
-function (n::DeterministicCNF)(x, p=n.p)
+function (n::DeterministicCNF)(x, p=n.p; sensealg=InterpolatingAdjoint(autojacvec=false))
```
ChrisRackauckas, line 74: this one is already handled in the kwargs of the DeterminsticCNF constructor.

src/ffjord.jl
```diff
@@ -225,26 +223,30 @@ function forward_ffjord(n::FFJORD, x, p=n.p, e=randn(eltype(x), size(x));
 end
 
 function backward_ffjord(n::FFJORD, n_samples, p=n.p, e=randn(eltype(n.model[1].weight), n_samples);
-                         regularize=false, monte_carlo=true, rng=nothing)
+                         regularize=false, monte_carlo=true, rng=nothing, sensealg=InterpolatingAdjoint())
```
ChrisRackauckas, line 226: this is already handled by n.kwargs

src/ffjord.jl
```diff
@@ -194,9 +193,8 @@ end
 (n::FFJORD)(args...; kwargs...) = forward_ffjord(n, args...; kwargs...)
 
 function forward_ffjord(n::FFJORD, x, p=n.p, e=randn(eltype(x), size(x));
-                        regularize=false, monte_carlo=true)
+                        regularize=false, monte_carlo=true, sensealg=InterpolatingAdjoint())
```
ChrisRackauckas, line 196: this is also already handled by n.kwargs.

prbzrg:
Oh, I thought passing `sensealg` make errors (which is not) and it is better to have default `sensealg` as a function argument with a default value.
Thanks for quick review!

ChrisRackauckas:
We want to use the standard default because it's a much smarter polyalgorithm. It'll outperform this in most cases.

PR #631, opened 2021-09-27
fix a typo

prbzrg:


ChrisRackauckas:
NeuralDAE is somewhat broken... the NeuralODEMM is the better DAE form 😅 . NeuralDAE in total needs a bit more research still.

PR #637, opened 2021-10-14
fix wrong variable in neural_ode_sciml.md

wkirgsn:
one certain variable name is orphaned and should be renamed accordingly.


ChrisRackauckas:
thanks

PR #651, opened 2021-11-16
use zero matrix each time in jacobian_fn

### LINKED ISSUE #650: Wrong Jacobian Matrix
I'm not sure, but I think `z` must be reset in each iteration of the loop
https://github.com/SciML/DiffEqFlux.jl/blob/46f88397eede901de0a954dce30c6d828ae7f977/src/ffjord.jl#L154-L164

```julia
julia> m = Dense(2, 2)
Dense(2, 2)         # 6 parameters

julia> r = rand(2, 5)
2×5 Matrix{Float64}:
 0.357079  0.647137  0.156209  0.181322  0.580713
 0.789372  0.171408  0.673694  0.711936  0.246555

julia> jacobian_fn(m, r)
2×2×5 Array{Float64, 3}:
[:, :, 1] =
 0.272913  0.219132
 1.26899   0.939335

[:, :, 2] =
 0.272913  0.219132
 1.26899   0.939335

[:, :, 3] =
 0.272913  0.219132
 1.26899   0.939335

[:, :, 4] =
 0.272913  0.219132
 1.26899   0.939335

[:, :, 5] =
 0.272913  0.219132
 1.26899   0.939335

julia> only(Zygote.jacobian(m, r))
10×10 Matrix{Float64}:
 0.272913  0.219132  0.0       0.0       0.0       0.0       0.0       0.0       0.0       0.0
 0.996074  0.720204  0.0       0.0       0.0       0.0       0.0       0.0       0.0       0.0
 0.0       0.0       0.272913  0.219132  0.0       0.0       0.0       0.0       0.0       0.0
 0.0       0.0       0.996074  0.720204  0.0       0.0       0.0       0.0       0.0       0.0
 0.0       0.0       0.0       0.0       0.272913  0.219132  0.0       0.0       0.0       0.0
 0.0       0.0       0.0       0.0       0.996074  0.720204  0.0       0.0       0.0       0.0
 0.0       0.0       0.0       0.0       0.0       0.0       0.272913  0.219132  0.0       0.0
 0.0       0.0       0.0       0.0       0.0       0.0       0.996074  0.720204  0.0       0.0
 0.0       0.0       0.0       0.0       0.0       0.0       0.0       0.0       0.272913  0.219132
 0.0       0.0       0.0       0.0       0.0       0.0       0.0       0.0       0.996074  0.720204

```

after change:
```julia
julia> function jacobian_fn(f, x::AbstractMatrix)
           y, back = Zygote.pullback(f, x)
           z = Zygote.@ignore similar(y)
           Zygote.@ignore fill!(z, zero(eltype(x)))
           vec = Zygote.Buffer(x, size(x, 1), size(x, 1), size(x, 2))
           for i in 1:size(y, 1)
               Zygote.@ignore z[i, :] .+= one(eltype(x))
               vec[i, :, :] = back(z)[1]
               # --- change
               z = Zygote.@ignore similar(y)
               Zygote.@ignore fill!(z, zero(eltype(x)))
               # --- change
           end
           copy(vec)
       end
jacobian_fn (generic function with 1 method)

julia> jacobian_fn(m, r)
2×2×5 Array{Float64, 3}:
[:, :, 1] =
 0.272913  0.219132
 0.996074  0.720204

[:, :, 2] =
 0.272913  0.219132
 0.996074  0.720204

[:, :, 3] =
 0.272913  0.219132
 0.996074  0.720204

[:, :, 4] =
 0.272913  0.219132
 0.996074  0.720204

[:, :, 5] =
 0.272913  0.219132
 0.996074  0.720204

```

prbzrg:
Fixes #650


src/ffjord.jl
```diff
@@ -157,8 +157,9 @@ function jacobian_fn(f, x::AbstractMatrix)
     Zygote.@ignore fill!(z, zero(eltype(x)))
     vec = Zygote.Buffer(x, size(x, 1), size(x, 1), size(x, 2))
     for i in 1:size(y, 1)
-        Zygote.@ignore z[i, :] .+= one(eltype(x))
-        vec[i, :, :] = back(z)[1]
+        z_ = copy(z)
```
ChrisRackauckas, line 160: it would be better to reset to zero I think.

src/ffjord.jl
```diff
@@ -159,6 +159,7 @@ function jacobian_fn(f, x::AbstractMatrix)
     for i in 1:size(y, 1)
         Zygote.@ignore z[i, :] .+= one(eltype(x))
         vec[i, :, :] = back(z)[1]
+        Zygote.@ignore fill!(z, zero(eltype(x)))
```
ChrisRackauckas, line 162: It probably doesn't matter much, but it's probably best to just zero the parts that changed, i.e. `Zygote.@ignore z[i, :] .= zero(eltype(x))`

src/ffjord.jl
prbzrg: Yes, I used `BenchmarkTools`, the results are:
```julia
# copy
  38.800 μs (291 allocations: 10.91 KiB)
# reset
  37.200 μs (291 allocations: 10.91 KiB)
```

src/ffjord.jl
```diff
@@ -159,6 +159,7 @@ function jacobian_fn(f, x::AbstractMatrix)
     for i in 1:size(y, 1)
         Zygote.@ignore z[i, :] .+= one(eltype(x))
```
ChrisRackauckas, line 160:
```suggestion
        Zygote.@ignore z[i, :] .= one(eltype(x))
```

both work but this is a bit more sane.

prbzrg: indeed

ChrisRackauckas:
Rebase to fix the CI.

ChrisRackauckas:
Ouch, this downstream failure is back? I'll need to take a look, but I thought it was tested...

ChrisRackauckas:
Oh,

https://github.com/SciML/DiffEqFlux.jl/runs/4224671354?check_suite_focus=true#step:6:120

 that's an old version. Gotta track that down. The latest versions are fine: https://github.com/SciML/DiffEqSensitivity.jl/runs/4588867599?check_suite_focus=true

ChrisRackauckas:
I rebased onto working tests.

ChrisRackauckas:
Seems like a real test failure?

prbzrg:
I think v1.7 change is causing the problem:

> The default random number generator has changed, so all random numbers will be different (even with the same seed) unless an explicit RNG object is used.

ChrisRackauckas:
Is it just a bad seed? It's not just v1.7 in isolation since the tests are passing on v1.7 without this

ChrisRackauckas:
Or are the tests incorrect?

prbzrg:
I think, it's better to test `FFJORD` against other methods like `kde` from `KernelDensity.jl`

prbzrg:
Or `ash` from `AverageShiftedHistograms.jl`

ChrisRackauckas:
So is the current test case incorrect, or does it just need a more stable training procedure?

prbzrg:
IMO, the training procedure is fine, maybe a hidden layer in neural network would improve the results.
What is the value `0.35` based on?

ChrisRackauckas:
That's just a regression test value. Maybe it can be safely bumped to 1.

ChrisRackauckas:
Seems to be a random thing. The losses are not very different in form from the passing case https://github.com/SciML/DiffEqFlux.jl/runs/4602555392?check_suite_focus=true

prbzrg:
I have done some experiment, unoptimized `FFJORD` make `totalvariation` `0.85`.

ChrisRackauckas:
Thanks for sticking with it. Was a hectic month.

PR #653, opened 2021-11-16
[DON'T MERGE] add DiffEqSensitivity upper bound

prbzrg:
Tests are failing, because of `DiffEqSensitivity@6.60.3`.
This PR demonstrates the simplest solution, but we should fix the actual issue.
What does make the problem?

https://github.com/SciML/DiffEqSensitivity.jl/compare/v6.60.2...v6.60.3


ChrisRackauckas:
I know what the problem is but just haven't had time to fix it. 

https://github.com/SciML/DiffEqSensitivity.jl/pull/506/files#diff-4c1a437283c1ed3aeb4267af6e255f3434d8ba2b62eb67192c482a6becd3192dR613

That breaks the gradient pullback. The real problem is that for some reason

https://github.com/SciML/DiffEqSensitivity.jl/commit/1f238d028d05ee0df9a82d697e141cfc8080ef97

that form of hcat also drops gradients.

ChrisRackauckas:
@frankschae if this is what is blocking you right now.

ChrisRackauckas:
@avik-pal if you have an idea.

ChrisRackauckas:
All is passing now.

PR #671, opened 2022-01-06
Preallocation Option for vjp calculation

Abhishek-1Bhatt:
Added precaching feature to avoid heap allocations on vjp calculation in reverse pass

### DIFF
--- a/src/fast_layers.jl
+++ b/src/fast_layers.jl
@@ -34,30 +34,35 @@ the layer and it defaults to true.
 Note that this function has specializations on `tanh` for a slightly faster
 adjoint with Zygote.
 """
-struct FastDense{F,F2} <: FastLayer
+struct FastDense{F,F2,C} <: FastLayer
   out::Int
   in::Int
   σ::F
   initial_params::F2
   bias::Bool
-  precache::Bool
-  cs :: NamedTuple
+  cache :: C
   function FastDense(in::Integer, out::Integer, σ = identity;
                  bias = true, precache=true, initW = Flux.glorot_uniform, initb = Flux.zeros32)
     temp = ((bias == false) ? vcat(vec(initW(out, in))) : vcat(vec(initW(out, in)),initb(out)))
     initial_params() = temp
-    cs = (
-      zbar = zeros(out, 1),
-      Wbar = zeros(out, in),
-      bbar = zeros(out, 1),
-      xbar = zeros(in, 1),
-      pbar = if bias == true
-          zeros((out*in)+out)
-        else
-          zeros(out*in)
-        end
-      )
-      new{typeof(σ), typeof(initial_params)}(out,in,σ,initial_params,bias,precache,cs)
+    if precache == true
+      cache = (
+        y = zeros(out),
+        r = zeros(out),
+        zbar = zeros(out, 1),
+        Wbar = zeros(out, in),
+        bbar = zeros(out, 1),
+        xbar = zeros(in, 1),
+        pbar = if bias == true
+            zeros((out*in)+out)
+          else
+            zeros(out*in)
+          end
+        )
+      else
+        cache = nothing
+      end
+      new{typeof(σ), typeof(initial_params), typeof(cache)}(out,in,σ,initial_params,bias,cache)
       # new{typeof(σ),typeof(initial_params)}(out,in,σ,initial_params,bias)
   end
 end
@@ -72,12 +77,21 @@ ZygoteRules.@adjoint function (f::FastDense)(x,p)
     W = reshape(@view(p[1:(f.out*f.in)]),f.out,f.in)
   end
 
-  if f.bias == true
-    b = p[(f.out*f.in + 1):end]
-    r = W*x .+ b
-    ifgpufree(b)
+  if f.cache == nothing
+    if f.bias == true
+      b = p[(f.out*f.in + 1):end]
+      r = W*x .+ b
+      ifgpufree(b)
+    else
+      r = W*x
+    end
+    y = f.σ.(r)
   else
-    r = W*x
+    mul!(f.cache.r, W, x); ###alloc
+    if f.bias == true
+      f.cache.r .+= p[(f.out*f.in + 1):end]
+    end
+    f.cache.y .= f.σ.(f.cache.r)
   end
 
   # if typeof(x) <: AbstractVector
@@ -89,33 +103,8 @@ ZygoteRules.@adjoint function (f::FastDense)(x,p)
   #   mul!(r,W,x,one(eltype(x)),one(eltype(x)))
   # end
 
-  y = f.σ.(r)
-
   function FastDense_adjoint(ȳ)
-    if(f.precache == true)
-      if typeof(f.σ) <: typeof(tanh)
-        f.cs.zbar = ȳ .* (1 .- y.^2)
-      elseif typeof(f.σ) <: typeof(identity)
-        f.cs.zbar = ȳ
-      else
-        f.cs.zbar = ȳ .* ForwardDiff.derivative.(f.σ,r)
-      end
-      f.cs.Wbar = f.cs.zbar * x'
-      f.cs.bbar = f.cs.zbar
-      f.cs.xbar = W' * f.cs.zbar
-      f.cs.pbar = if f.bias == true
-          tmp = typeof(f.cs.bbar) <: AbstractVector ? #how to find if bbar is AbstractVector and allocate its shape and size
-                           vec(vcat(vec(f.cs.Wbar),f.cs.bbar)) :
-                           vec(vcat(vec(f.cs.Wbar),sum(f.cs.bbar,dims=2)))
-          ifgpufree(f.cs.bbar)
-          tmp
-      else
-          vec(f.cs.Wbar)
-      end
-      ifgpufree(f.cs.Wbar)
-      ifgpufree(r)
-      nothing,f.cs.xbar,f.cs.pbar
-    else
+    if f.cache == nothing
       if typeof(f.σ) <: typeof(tanh)
         zbar = ȳ .* (1 .- y.^2)
       elseif typeof(f.σ) <: typeof(identity)
@@ -138,9 +127,35 @@ ZygoteRules.@adjoint function (f::FastDense)(x,p)
       ifgpufree(Wbar)
       ifgpufree(r)
       nothing,xbar,pbar
+    else
+      if typeof(f.σ) <: typeof(tanh)
+        f.cache.zbar .= ȳ .* (1 .- f.cache.y.^2)
+      elseif typeof(f.σ) <: typeof(identity)
+        f.cache.zbar .= ȳ
+      else
+        f.cache.zbar .= ȳ .* ForwardDiff.derivative.(f.σ,f.cache.r)
+      end
+      # f.cache.Wbar .= f.cache.zbar * x'
+      mul!(f.cache.Wbar, f.cache.zbar, x');
+      f.cache.bbar .= f.cache.zbar
+      # f.cache.xbar .= W' * f.cache.zbar
+      mul!(f.cache.xbar, W', f.cache.zbar);#alloc
+      f.cache.pbar .= if f.bias == true
+          tmp = typeof(f.cache.bbar) <: AbstractVector ?
+                           vec(vcat(vec(f.cache.Wbar),f.cache.bbar)) :
+                           vec(vcat(vec(f.cache.Wbar),sum(f.cache.bbar,dims=2)))
+          tmp
+      else
+          vec(f.cache.Wbar)
+      end
+      nothing,f.cache.xbar,f.cache.pbar
     end
   end
-  y,FastDense_adjoint
+  if f.cache == nothing
+    y,FastDense_adjoint
+  else
+    f.cache.y, FastDense_adjoint
+  end
 end
 paramlength(f::FastDense) = f.out*(f.in+f.bias)
 initial_params(f::FastDense) = f.initial_params()
--- a/src/fast_layers.jl
+++ b/src/fast_layers.jl
@@ -24,12 +24,14 @@ initial_params(c::FastChain) = vcat(initial_params.(c.layers)...)
 
 """
 FastDense(in,out,activation=identity;
-          bias = true ,initW = Flux.glorot_uniform, initb = Flux.zeros32)
+          bias = true, precache = false ,initW = Flux.glorot_uniform, initb = Flux.zeros32)
 
 A Dense layer `activation.(W*x + b)` with input size `in` and output size `out`.
 The `activation` function defaults to `identity`, meaning the layer is an affine
 function. Initial parameters are taken to match `Flux.Dense`. 'bias' represents b in
-the layer and it defaults to true.
+the layer and it defaults to true.'precache' is used to preallocate memory for the
+intermediate variables calculated during each pass. This avoids heap allocations
+in each pass which would otherwise slow down the computation, it defaults to false.
 
 Note that this function has specializations on `tanh` for a slightly faster
 adjoint with Zygote.
@@ -39,20 +41,21 @@ struct FastDense{F,F2,C} <: FastLayer
   in::Int
   σ::F
   initial_params::F2
-  bias::Bool
   cache :: C
+  bias::Bool
+  numcols::Int64
   function FastDense(in::Integer, out::Integer, σ = identity;
-                 bias = true, precache=true, initW = Flux.glorot_uniform, initb = Flux.zeros32)
+                 bias = true, numcols=1, precache=false, initW = Flux.glorot_uniform, initb = Flux.zeros32)
     temp = ((bias == false) ? vcat(vec(initW(out, in))) : vcat(vec(initW(out, in)),initb(out)))
     initial_params() = temp
     if precache == true
       cache = (
-        y = zeros(out),
-        r = zeros(out),
-        zbar = zeros(out, 1),
+        W = zeros(out, in),
+        y = zeros(out, numcols),
+        r = zeros(out, numcols),
+        zbar = zeros(out, numcols),
         Wbar = zeros(out, in),
-        bbar = zeros(out, 1),
-        xbar = zeros(in, 1),
+        xbar = zeros(in, numcols),
         pbar = if bias == true
             zeros((out*in)+out)
           else
@@ -62,7 +65,7 @@ struct FastDense{F,F2,C} <: FastLayer
       else
         cache = nothing
       end
-      new{typeof(σ), typeof(initial_params), typeof(cache)}(out,in,σ,initial_params,bias,cache)
+      new{typeof(σ), typeof(initial_params), typeof(cache)}(out,in,σ,initial_params,cache,bias,numcols)
       # new{typeof(σ),typeof(initial_params)}(out,in,σ,initial_params,bias)
   end
 end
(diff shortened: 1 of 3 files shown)


src/fast_layers.jl
```diff
@@ -40,11 +40,25 @@ struct FastDense{F,F2} <: FastLayer
   σ::F
   initial_params::F2
   bias::Bool
+  precache::Bool
```
ChrisRackauckas, line 43: there's no need to store this. If `precache=false`, then just store `nothing` and check for that in the function.

src/fast_layers.jl
```diff
@@ -40,11 +40,25 @@ struct FastDense{F,F2} <: FastLayer
   σ::F
   initial_params::F2
   bias::Bool
+  precache::Bool
+  cs :: NamedTuple
```
ChrisRackauckas, line 44: Use informative variable names. `cache` instead of `cs`.

Also, this is not type stable, instead parameterize it `::C`.

src/fast_layers.jl
```diff
@@ -78,28 +92,53 @@ ZygoteRules.@adjoint function (f::FastDense)(x,p)
   y = f.σ.(r)
```
ChrisRackauckas, line 92: you missed this allocation and some before.

src/fast_layers.jl
```diff
@@ -78,28 +92,53 @@ ZygoteRules.@adjoint function (f::FastDense)(x,p)
...
+        f.cs.zbar = ȳ .* (1 .- y.^2)
+      elseif typeof(f.σ) <: typeof(identity)
+        f.cs.zbar = ȳ
+      else
+        f.cs.zbar = ȳ .* ForwardDiff.derivative.(f.σ,r)
+      end
+      f.cs.Wbar = f.cs.zbar * x'
+      f.cs.bbar = f.cs.zbar
+      f.cs.xbar = W' * f.cs.zbar
+      f.cs.pbar = if f.bias == true
```
ChrisRackauckas, line 106: These lines all still allocate. They need `.=`, `mul!`, etc.

src/fast_layers.jl
```diff
@@ -78,28 +92,53 @@ ZygoteRules.@adjoint function (f::FastDense)(x,p)
...
+          tmp = typeof(f.cs.bbar) <: AbstractVector ? #how to find if bbar is AbstractVector and allocate its shape and size
+                           vec(vcat(vec(f.cs.Wbar),f.cs.bbar)) :
+                           vec(vcat(vec(f.cs.Wbar),sum(f.cs.bbar,dims=2)))
+          ifgpufree(f.cs.bbar)
```
ChrisRackauckas, line 110: If it's in the cache, then don't free it.

src/fast_layers.jl
```diff
@@ -78,28 +92,53 @@ ZygoteRules.@adjoint function (f::FastDense)(x,p)
...
+        f.cs.zbar = ȳ
+      else
+        f.cs.zbar = ȳ .* ForwardDiff.derivative.(f.σ,r)
+      end
+      f.cs.Wbar = f.cs.zbar * x'
+      f.cs.bbar = f.cs.zbar
+      f.cs.xbar = W' * f.cs.zbar
+      f.cs.pbar = if f.bias == true
+          tmp = typeof(f.cs.bbar) <: AbstractVector ? #how to find if bbar is AbstractVector and allocate its shape and size
+                           vec(vcat(vec(f.cs.Wbar),f.cs.bbar)) :
```
ChrisRackauckas, line 108: These are allocating statements.

ChrisRackauckas:
It's a start but still has a long way to go. Need to test gradient accuracy in https://github.com/SciML/DiffEqFlux.jl/blob/master/test/fast_layers.jl (and inaccuracy in second calls), and should add a test for its usage in neural ODEs in https://github.com/SciML/DiffEqFlux.jl/blob/master/test/neural_de.jl and https://github.com/SciML/DiffEqFlux.jl/blob/master/test/neural_de_gpu.jl

Abhishek-1Bhatt:
Thanks for the feedback : ), I'll fix these issues

ChrisRackauckas:
What's the status here?

Abhishek-1Bhatt:
Just pushed the required updates for allowing matrix inputs. It takes views when number of columns in input is less than pre specified number, otherwise everything is done with the preallocated buffers with full size(1 by default).

test/neural_de.jl
```diff
@@ -90,6 +91,38 @@ grads = Zygote.gradient(()->sum(node(xs)),Flux.params(xs,node))
 goodgrad2 = grads[node.p]
 @test goodgrad ≈ goodgrad2 # Make sure adjoint overloads are correct
 
+node = NeuralODE(fastcdudt,tspan,Tsit5(),save_everystep=false,save_start=false)
+grads = Zygote.gradient(()->sum(node(x)),Flux.params(x,node))
+@test ! iszero(grads[x])
+@test ! iszero(grads[node.p])
+
```
ChrisRackauckas, line 98: We should test that these gradients match the non-caching one.

test/fast_layers.jl
```diff
@@ -38,6 +44,12 @@ fsgrad = Flux.Zygote.gradient((x,p)->sum(fs(x,p)),x,pd)
 @test fdgrad[1] ≈ fsgrad[1]
 @test fdgrad[2] ≈ fsgrad[2] rtol=1e-5
 
+fdcgrad = Flux.Zygote.gradient((x,p)->sum(fdc(x,p)),x,pd)
+@test fdgrad[1] ≈ fdcgrad[1]
+@test fdgrad[2] ≈ fdcgrad[2] rtol=1e-5
```
ChrisRackauckas, line 49: why so high of a tolerance? Seems like an issue?

test/fast_layers.jl
```diff
@@ -38,6 +44,12 @@ fsgrad = Flux.Zygote.gradient((x,p)->sum(fs(x,p)),x,pd)
 @test fdgrad[1] ≈ fsgrad[1]
 @test fdgrad[2] ≈ fsgrad[2] rtol=1e-5
 
+fdcgrad = Flux.Zygote.gradient((x,p)->sum(fdc(x,p)),x,pd)
+@test fdgrad[1] ≈ fdcgrad[1]
+@test fdgrad[2] ≈ fdcgrad[2] rtol=1e-5
+@allocated fdc(x, pd);
+@test @allocated fdc(x, pd) == 1024
```
ChrisRackauckas, line 51: What are these allocations from?

src/fast_layers.jl
```diff
@@ -24,84 +24,160 @@ initial_params(c::FastChain) = vcat(initial_params.(c.layers)...)
...
-  y = f.σ.(r)
-
-  function FastDense_adjoint(ȳ)
-    if typeof(f.σ) <: typeof(tanh)
-      zbar = ȳ .* (1 .- y.^2)
-    elseif typeof(f.σ) <: typeof(identity)
-      zbar = ȳ
+    cols = length(size(x)) == 1 ? 1 : size(x)[2]
+    if !isgpu(p)
+      f.cache.W .= @view p[reshape(1:(f.out*f.in),f.out,f.in)]
```
ChrisRackauckas, line 94: why is this cached?

src/fast_layers.jl
```diff
@@ -24,84 +24,160 @@ initial_params(c::FastChain) = vcat(initial_params.(c.layers)...)
...
+      end
+      new{typeof(σ), typeof(initial_params), typeof(cache)}(out,in,σ,initial_params,cache,bias,numcols)
+      # new{typeof(σ),typeof(initial_params)}(out,in,σ,initial_params,bias)
   end
 end
 
 # (f::FastDense)(x,p) = f.σ.(reshape(uview(p,1:(f.out*f.in)),f.out,f.in)*x .+ uview(p,(f.out*f.in+1):lastindex(p)))
 (f::FastDense)(x,p) = ((f.bias == true) ? (f.σ.(reshape(p[1:(f.out*f.in)],f.out,f.in)*x .+ p[(f.out*f.in+1):end])) : (f.σ.(reshape(p[1:(f.out*f.in)],f.out,f.in)*x)))
 
 ZygoteRules.@adjoint function (f::FastDense)(x,p)
```
ChrisRackauckas, line 76: Make this two separate dispatches, one for `x::AbstractVector` and one for `x::AbstractMatrix`. That will make it a lot simpler.

src/fast_layers.jl
Abhishek-1Bhatt: Did this just to avoid the pointer allocation from taking @view. Would you prefer it uncached?

src/fast_layers.jl
```diff
@@ -24,84 +24,160 @@ initial_params(c::FastChain) = vcat(initial_params.(c.layers)...)
...
-  # end
-
-  y = f.σ.(r)
-
-  function FastDense_adjoint(ȳ)
-    if typeof(f.σ) <: typeof(tanh)
-      zbar = ȳ .* (1 .- y.^2)
-    elseif typeof(f.σ) <: typeof(identity)
-      zbar = ȳ
+    cols = length(size(x)) == 1 ? 1 : size(x)[2]
```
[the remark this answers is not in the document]
Abhishek-1Bhatt: This would cause error when x::AbstractVector but numcols>1, although with separated dispatches as you suggested below it might not occur

test/fast_layers.jl
Abhishek-1Bhatt: would 1e-9 be ok, any specific value?

test/neural_de.jl
Abhishek-1Bhatt: for each test case or only this one here?

src/fast_layers.jl
Abhishek-1Bhatt: if uncached it causes further allocations when its transpose is taken [here](https://github.com/Abhishek-1Bhatt/DiffEqFlux.jl/blob/896fc48f9a82711a9ced560f5d543cbedfdcbfed/src/fast_layers.jl#L149)

ChrisRackauckas: Interesting, view of a transpose allocates?

ChrisRackauckas:
Looks like a real test failure? https://github.com/SciML/DiffEqFlux.jl/runs/4918538935?check_suite_focus=true#step:6:499

test/fast_layers.jl
ChrisRackauckas: 1e-9 is fine. For something that's just changing to no allocs, I would expect it to pass at like `1e-12` at least.

Abhishek-1Bhatt:
Could calculating with W and b as `Float32`s be ruining type inference for r & y?
I called @code_warntype like so:
`fdc = FastDense(2, 50, tanh, numcols=1,precache=true)
pd = initial_params(fdc)
x = rand(2)
@code_warntype Zygote.pullback(fdc,x,pd)`

Abhishek-1Bhatt:
with Float32 it return
Screenshot (24)
s

Abhishek-1Bhatt:
and with W and b wrapped with `Float64()` we get
Screenshot (26)

Abhishek-1Bhatt:
plus the profiler output was also showing allocations for r and y, which is the reason I did the above code_warntype calls

Abhishek-1Bhatt:
[This](https://github.com/Abhishek-1Bhatt/DiffEqFlux.jl/blob/5ff2129de1597372bbf811e031ef66e38c4d6381/src/fast_layers.jl#L228) return is causing the [allocations](https://github.com/Abhishek-1Bhatt/DiffEqFlux.jl/blob/5ff2129de1597372bbf811e031ef66e38c4d6381/test/fast_layers.jl#L51) here. Is there any workaround for this?

ChrisRackauckas:
You can have a preallocated vector for the return that you write into. Indeed it seems to instantiate the view.

Abhishek-1Bhatt:
Making y a preallocated vector reduced some allocations but the Fastdense_adjoint that is also being returned here seems to be the main problem.

 Just for experimenting I returned `nothing, y` , i.e., nothing for y and y instead of FastDense_adjoint and with the following @allocated call it returned just 176, which suggests FastDense_adjoint allocating

y is just a placeholder here because if nothing is returned in place of FastDense_adjoint then Zygote.pullback will error out.

WhatsApp Image 2022-02-01 at 2 08 30 PM

src/fast_layers.jl
```diff
@@ -24,85 +24,212 @@ initial_params(c::FastChain) = vcat(initial_params.(c.layers)...)
...
+      else
+        vec(f.cache.Wbar)
+      end
+      nothing,@view(f.cache.xbar[:,1:f.cache.cols[1]]),f.cache.pbar
+    end
+  end
+  if typeof(f.cache) <: Nothing
+    y,FastDense_adjoint
+  else
+    @view(f.cache.y[:,1:f.cache.cols[1]]),FastDense_adjoint
```
ChrisRackauckas, line 229: have an `out` array that you write into and then return, instead of returning a view which will allocate the mutable struct of the view itself. That's probably the 176.

src/fast_layers.jl
```diff
@@ -24,85 +24,212 @@ initial_params(c::FastChain) = vcat(initial_params.(c.layers)...)
...
+        vec(f.cache.Wbar)
+      end
+      nothing,@view(f.cache.xbar[:,1]),f.cache.pbar
+    end
+  end
+  if typeof(f.cache) <: Nothing
+    y,FastDense_adjoint
   else
-    r = W*x
+    f.cache.yvec,FastDense_adjoint
```
Abhishek-1Bhatt, line 150: I am talking about this one. After replacing the view with a preallocated array the allocations decrease by ~400 bytes but it still allocates ~1400, which disappears when FastDense_adjoint is removed from above, although we can't compute if its removed.

ChrisRackauckas: oh yes that closure will need to allocate. Don't worry about that for now. If that's all that's left, then we're good. Let's try to get this finished, merged, and then talk about what to do here.

src/fast_layers.jl
Abhishek-1Bhatt: can we do anything for this view without the confirmation that numcols will equal cols?

Abhishek-1Bhatt: can we do anything for this view without the confirmation that numcols will equal cols?

Abhishek-1Bhatt: we can return the whole y if numcols equals cols and if cols is lesser, the view

test/fast_layers.jl
```diff
@@ -38,6 +44,12 @@ fsgrad = Flux.Zygote.gradient((x,p)->sum(fs(x,p)),x,pd)
 @test fdgrad[1] ≈ fsgrad[1]
 @test fdgrad[2] ≈ fsgrad[2] rtol=1e-5
 
+fdcgrad = Flux.Zygote.gradient((x,p)->sum(fdc(x,p)),x,pd)
+@test fdgrad[1] ≈ fdcgrad[1]
+@test fdgrad[2] ≈ fdcgrad[2] rtol=1e-12
+@allocated(Zygote.pullback(fdc, x, pd))
+@test @allocated(Zygote.pullback(fdc, x, pd)) < 1450
```
ChrisRackauckas, line 51: decrease?

test/fast_layers.jl
```diff
@@ -38,6 +44,12 @@ fsgrad = Flux.Zygote.gradient((x,p)->sum(fs(x,p)),x,pd)
 @test fdgrad[1] ≈ fsgrad[1]
 @test fdgrad[2] ≈ fsgrad[2] rtol=1e-5
 
+fdcgrad = Flux.Zygote.gradient((x,p)->sum(fdc(x,p)),x,pd)
+@test fdgrad[1] ≈ fdcgrad[1]
```
ChrisRackauckas, line 48: lower tolerance

test/neural_de.jl
```diff
@@ -59,9 +60,9 @@ grads = Zygote.gradient(()->sum(node(xs)),Flux.params(xs,node))
 @info "Test some fast layers"
 
 node = NeuralODE(fastdudt,tspan,Tsit5(),save_everystep=false,save_start=false)
-grads = Zygote.gradient(()->sum(node(x)),Flux.params(x,node))
-@test ! iszero(grads[x])
-@test ! iszero(grads[node.p])
+gradsnc = Zygote.gradient(()->sum(node(x)),Flux.params(x,node))
+@test ! iszero(gradsnc[x])
+@test ! iszero(gradsnc[node.p])
```
ChrisRackauckas, line 65: check that this matches the one without caching to a low tolerance. Set the ODE solver tolerance low for this test.

test/neural_de.jl
```diff
@@ -90,6 +91,40 @@ grads = Zygote.gradient(()->sum(node(xs)),Flux.params(xs,node))
 goodgrad2 = grads[node.p]
 @test goodgrad ≈ goodgrad2 # Make sure adjoint overloads are correct
 
+node = NeuralODE(fastcdudt,tspan,Tsit5(),save_everystep=false,save_start=false)
+gradsc = Zygote.gradient(()->sum(node(x)),Flux.params(x,node))
+@test ! iszero(gradsc[x])
+@test ! iszero(gradsc[node.p])
+@test gradsnc[x] ≈ gradsc[x]
```
ChrisRackauckas, line 98: tolerance on here?

test/neural_de.jl
```diff
@@ -90,6 +91,40 @@ grads = Zygote.gradient(()->sum(node(xs)),Flux.params(xs,node))
...
+node = NeuralODE(fastcdudt,tspan,Tsit5,save_everystep=false,save_start=false, sensealg=BacksolveAdjoint(),p=p)
+grads = Zygote.gradient(()->sum(node(x)),Flux.params(x,node))
+@test ! iszero(grads[x])
+@test ! iszero(grads[node.p])
+
+grads = Zygote.gradient(()->sum(node(xs)),Flux.params(xs,node))
+@test !iszero(grads[xs])
+@test ! iszero(grads[node.p])
+goodgrad2 = grads[node.p]
+@test goodgrad ≈ goodgrad2
```
ChrisRackauckas, line 126: etc

test/neural_de.jl
```diff
@@ -154,6 +189,33 @@ grads = Zygote.gradient(()->sum(node(x)),Flux.params(x,node))
...
+    @test_broken ! iszero(grads[node.p])
+
+    node = NeuralODE(fastcdudt,tspan,Tsit5(),saveat=0.1)
+    grads = Zygote.gradient(()->sum(node(x)),Flux.params(x,node))
+    @test ! iszero(grads[x])
+    @test ! iszero(grads[node.p])
+
+    @test_broken grads = Zygote.gradient(()->sum(node(xs)),Flux.params(xs,node)) isa Tuple
+    @test_broken ! iszero(grads[xs])
+    @test_broken ! iszero(grads[node.p])
```
ChrisRackauckas, line 218: It would be easier to read if the cached test is paired right next to the uncached test

test/neural_de.jl
```diff
@@ -244,8 +334,21 @@ grads = Zygote.gradient(()->sum(sode(x)),Flux.params(x,sode))
 @test ! iszero(grads[sode.p])
 @test ! iszero(grads[sode.p][end])
 
+sode = NeuralDSDE(fastcdudt,fastcdudt2,(0.0f0,.1f0),SOSRI(),saveat=0.0:0.01:0.1)
```
ChrisRackauckas, line 337: if you set RNG seeds, do these gradients match the uncached versions? That's a good test.

test/neural_de.jl
```diff
@@ -295,3 +410,15 @@ grads = Zygote.gradient(()->sum(dode(x)),Flux.params(x,dode))
...
+NeuralCDDE(fastcddudt,(0.0f0,2.0f0),(p,t)->zero(x),(1f-1,2f-1),MethodOfSteps(Tsit5()),saveat=0.1)(x)
+dode = NeuralCDDE(fastcddudt,(0.0f0,2.0f0),(p,t)->zero(x),(1f-1,2f-1),MethodOfSteps(Tsit5()),saveat=0.0:0.1:2.0)
+
+grads = Zygote.gradient(()->sum(dode(x)),Flux.params(x,dode))
+@test ! iszero(grads[x])
+@test ! iszero(grads[dode.p])
+
+@test_broken grads = Zygote.gradient(()->sum(dode(xs)),Flux.params(xs,dode)) isa Tuple
+@test_broken ! iszero(grads[xs])
+@test ! iszero(grads[dode.p])
```
ChrisRackauckas, line 424: Check against the uncached.

test/fast_layers.jl
Abhishek-1Bhatt: Now this 1376 is due to the closure(FastDense_adjoint) which we can't deal with right now

ChrisRackauckas: Yup okay, that's good. Tidy up the rest of the tests and I think it's good to merge.

ChrisRackauckas (approved):

ChrisRackauckas:
Test failure

Abhishek-1Bhatt:
On a simple adjoint calculation we get ~2x speedup with precaching
WhatsApp Image 2022-02-05 at 11 06 01 PM

PR #678, opened 2022-01-25
Update and restructuring of docs on `autojacvec` options

JordiBolibar:
I have restructured that section in order to make it more straightforward to read. I have also added a small comment regarding `ZygoteVJP()`, warning about the non mutability and how this impacts the performance during the solve.

I also think this could be a good chance to further improve the comments on each type of `autojacvec`, following the recent discussion on Discourse and Slack. Particularly for `TrackerVJP`, `true` and `false`.


ChrisRackauckas:
Thanks!

PR #688, opened 2022-03-03
add prediction-error method tutorial

baggepinnen:
This PR adds a tutorial on using the prediction-error method (PEM) for parameter estimation. 

I noticed that none of the tutorials are run in CI, and images are hardcoded in. There are currently no images inserted into this tutorial, and I'm wondering if
- I should save images somewhere like for the other tutorials. If so, how do I go about that?
- Make Documenter.jl run and produce the images?

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -31,6 +31,7 @@ makedocs(
         "Training Techniques" => Any[
             "examples/multiple_shooting.md",
             "examples/local_minima.md",
+            "examples/prediction_error_method.md",
             "examples/divergence.md",
             "examples/multiple_nn.md",
             "examples/data_parallel.md",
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
> I should save images somewhere like for the other tutorials. If so, how do I go about that?

Just plop them in an issue or PR and use the link.

> Make Documenter.jl run and produce the images?

We could try to turn on doctesting, though these docs will need the buildkite 😅 . Most of the docs are part of the tests though.

baggepinnen:
img1
img2
img3
img4
img5

baggepinnen:
The trick with the images worked well, I'll remember that :) 
The page can be viewed in mostly rendered form here https://github.com/SciML/DiffEqFlux.jl/blob/c55b8793e91a931afdb2a5b6bc29b2178f56fbbe/docs/src/examples/prediction_error_method.md
This PR should be ready now

ChrisRackauckas:
Wow, this is a cool trick. Awesome!

PR #697, opened 2022-03-31
Generalize `input_params` to use `Flux.destructure`

DhairyaLGandhi:
Ref https://github.com/FluxML/DaggerFlux.jl/pull/17, this should get https://neuralpde.sciml.ai/dev/pinn/2D/ to run with DaggerFlux.

GitHub is showing a weird diff. It was an easy change so I did this with the web UI, which I guess is the culprit.

### DIFF
--- a/src/fast_layers.jl
+++ b/src/fast_layers.jl
@@ -1,8 +1,7 @@
 abstract type FastLayer <: Function end
 
 paramlength(f) = 0
-initial_params(f) = Float32[]
-initial_params(f::Chain) = Flux.destructure(f)[1]
+initial_params(f) = Flux.destructure(f)[1]
 
 struct FastChain{T<:Tuple} <: FastLayer
   layers::T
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
The only substantive change I see is unicode changes?

DhairyaLGandhi:
The unicode stuff wasn't intentional and caused by the GitHub web UI. The change is that I've replaced the generic `initial_params(f)` method with the one for `Chain` to accommodate any other chain-likes such as DaggerChain automatically. Without a method that DaggerChain can go to, it ends up returning an empty array which I guess was there for type stability.

ChrisRackauckas:
It looks like this causes test failures? https://github.com/SciML/DiffEqFlux.jl/runs/5771438406?check_suite_focus=true#step:6:460

DhairyaLGandhi:
Right, the error suggests that something tried access the first element of an empty array. This should have failed even when the fallback method got hit iiuc?

PR #705, opened 2022-05-03
Keep support for `cb` to avoid breakage

Vaibhavdixit02:
@ChrisRackauckas


ChrisRackauckas:
Oh oops, not handling this was a mistake. Can you merge and tag as a patch?

Vaibhavdixit02:
yes I will do that, yeah it was a mistake 🤦

ChrisRackauckas:
The GPU failure looks like a Flux GPU not early freeing thing, which the lower memory use of Lux.jl might help with 😅

PR #711, opened 2022-05-30
swapped sciml_train calls with Optimization.solve()

Abhishek-1Bhatt:
[SciML/DiffEqSensitivity.jl/issues/582](https://github.com/SciML/DiffEqSensitivity.jl/issues/582)


Vaibhavdixit02 (commented):
You'd need to load `GalacticOptimJL` or `GalacticFlux` as well

Abhishek-1Bhatt:
I'll add `GalacticOptimJL` to the test module's environment then(Project.toml) ?

Vaibhavdixit02:
If you plan to update the tests, then yeah. But I actually meant that you should also load these packages in the docs.

docs/src/examples/collocation.md
```diff
@@ -5,7 +5,7 @@ pretraining the neural network against a smoothed collocation of the
 data. First the example and then an explanation.
 
 ```julia
-using DiffEqFlux, DifferentialEquations, Plots
+using DiffEqFlux, DifferentialEquations, GalacticOptim, GalacticOptimJL, Plots
```
Vaibhavdixit02, line 8:
```suggestion
using DiffEqFlux, DifferentialEquations, GalacticOptim, GalacticFlux, Plots
```

docs/src/examples/multiple_shooting.md
```diff
@@ -18,7 +18,7 @@ high penalties in case the solver predicts discontinuous values.
 The following is a working demo, using Multiple Shooting
 
 ```julia
-using DiffEqFlux, DifferentialEquations, Plots
+using DiffEqFlux, GalacticOptim, , GalacticOptimJL, DifferentialEquations, Plots
```
Vaibhavdixit02, line 21:
```suggestion
using DiffEqFlux, GalacticOptim, GalacticPolyalgorithm, DifferentialEquations, Plots
```

docs/src/examples/normalizing_flows.md
```diff
@@ -8,7 +8,7 @@ Before getting to the explanation, here's some code to start with. We will
 follow a full explanation of the definition and training process:
 
 ```julia
-using DiffEqFlux, DifferentialEquations, GalacticOptim, Distributions
+using DiffEqFlux, DifferentialEquations, GalacticOptim, GalacticOptimJL, Distributions
```
Vaibhavdixit02, line 11:
```suggestion
using DiffEqFlux, DifferentialEquations, GalacticOptim, GalacticFlux, GalacticOptimJL, Distributions
```

docs/src/examples/normalizing_flows.md
```diff
@@ -27,8 +27,8 @@ function loss(θ)
 end
 
 adtype = GalacticOptim.AutoZygote()
-res1 = DiffEqFlux.sciml_train(loss, ffjord_mdl.p, ADAM(0.1), adtype; maxiters=100)
-res2 = DiffEqFlux.sciml_train(loss, res1.u, LBFGS(), adtype; allow_f_increases=false)
+res1 = GalacticOptim.solve(loss, ffjord_mdl.p, ADAM(0.1), adtype; maxiters=100)
+res2 = GalacticOptim(loss, res1.u, LBFGS(), adtype; allow_f_increases=false)
```
Vaibhavdixit02, line 31:
```suggestion
res2 = GalacticOptim.solve(loss, res1.u, Optim.LBFGS(), adtype; allow_f_increases=false)
```

docs/src/examples/normalizing_flows.md
```diff
@@ -47,7 +47,7 @@ new_data = rand(ffjord_dist, 100)
 We can use DiffEqFlux.jl to define, train and output the densities computed by CNF layers. In the same way as a neural ODE, the layer takes a neural network that defines its derivative function (see [1] for a reference). A possible way to define a CNF layer, would be:
 
 ```julia
-using DiffEqFlux, DifferentialEquations, GalacticOptim, Distributions
+using DiffEqFlux, DifferentialEquations, GalacticOptim, GalacticOptimJL, Distributions
```
Vaibhavdixit02, line 50:
```suggestion
using DiffEqFlux, DifferentialEquations, GalacticOptim, GalacticFlux, GalacticOptimJL, Distributions
```

Vaibhavdixit02:
You'd want to run the docs' examples locally since we don't have doctests, to make sure they are correct

ChrisRackauckas:
Yes, it should never ever show up in user code.

ChrisRackauckas:
Indeed, I don't understand how instantiate_function ended up here. Is there a single tutorial or documentation in any of Optimization.jl or GalacticOptim.jl that mentions it? I don't think there is, right?

Abhishek-1Bhatt:
Yeah, this is it, I couldn't remember earlier. But still I should have checked the docs of Optimization.jl directly.

ChrisRackauckas:
ahh, that makes sense. Okay, I eliminated it from https://github.com/SciML/DiffEqFlux.jl/pull/712

ChrisRackauckas:
Can we enable doctests with this one? Follow https://github.com/SciML/ModelingToolkit.jl/pull/1572 as an example.

src/ffjord.jl
```diff
@@ -169,7 +188,12 @@ _trace_batched(x::AbstractArray{T, 3}) where T =
 
 function ffjord(u, p, t, re, e;
                 regularize=false, monte_carlo=true)
-    m = re(p)
+    # m = re(p)
+    if isnothing(re)
+        m = p
```
Abhishek-1Bhatt, line 193: FFJORD still needs some work because here m will be a `ComponentVector` which is not callable when doing  `mz = m(z)` which throws an error. Not sure how to resolve this.

src/neural_de.jl
```diff
@@ -36,29 +37,40 @@ References:
...
+    function NeuralODE(model::FastChain,tspan,args...;p = initial_params(model),st=nothing,kwargs...)
         re = nothing
-        new{typeof(model),typeof(p),typeof(re),
+        new{typeof(model),typeof(p),typeof(st),typeof(re),
             typeof(tspan),typeof(args),typeof(kwargs)}(
-            model,p,re,tspan,args,kwargs)
+            model,p,st,re,tspan,args,kwargs)
+    end
+
+    function NeuralODE(model::Lux.Chain,tspan,args...;p=nothing,st=nothing,kwargs...)
```
avik-pal, line 66:
```suggestion
    function NeuralODE(model::Lux.AbstractExplicitLayer,tspan,args...;p=nothing,st=nothing,kwargs...)
```

src/neural_de.jl
```diff
@@ -78,6 +90,14 @@ function (n::NeuralODE{M})(x,p=n.p) where {M<:FastChain}
     solve(prob,n.args...;sensealg=sense,n.kwargs...)
 end
 
+function (n::NeuralODE{M})(x,p=n.p,st=n.st) where {M<:Lux.Chain}
+  dudt_(u,p,t) = n.model(u,p,st)[1]
```
avik-pal, line 94: This might lead to unexpected results. You want to capture state changes. Something like
```julia
function (n::NeuralODE)(x, ps, st)
    function dudt(u, p, t)
        u_, st = n.model(u, p, st)
        return u_
    end
    prob = ODEProblem{false}(ODEFunction{false}(dudt), x, n.tspan, ps)
    return solve(prob, n.solver; sensealg=n.sensealg, n.kwargs...), st
end
```

src/neural_de.jl
```diff
@@ -78,6 +90,14 @@ function (n::NeuralODE{M})(x,p=n.p) where {M<:FastChain}
     solve(prob,n.args...;sensealg=sense,n.kwargs...)
 end
 
+function (n::NeuralODE{M})(x,p=n.p,st=n.st) where {M<:Lux.Chain}
```
ChrisRackauckas, line 93:
```suggestion
function (n::NeuralODE{M})(x,p=n.p,st=n.st) where {M<:Lux.AbstractExplicitLayer}
```

src/ffjord.jl
```diff
@@ -119,18 +121,35 @@ References:
...
+        end
+        new{typeof(model),typeof(p),typeof(st),typeof(re),
+          typeof(basedist),typeof(tspan),typeof(args),typeof(kwargs)}(
+          model,p,st,re,basedist,tspan,args,kwargs)
+    end
 end
 
 function FFJORD(model, tspan, args...;
-                p::P=nothing, basedist::D=nothing, kwargs...) where {P <: Union{AbstractVector{<: AbstractFloat}, Nothing}, RE <: Function, D <: Union{Distribution, Nothing}}
+                p::P=nothing, st::ST=nothing, basedist::D=nothing, kwargs...) where {P <: Union{AbstractVector{<: AbstractFloat}, Nothing}, ST <: Nothing, RE <: Function, D <: Union{Distribution, Nothing}}
```
ChrisRackauckas, line 152: No, it's not always `<: Nothing`, it just defaults to that. Get right of the `::ST` here.

src/ffjord.jl
```diff
@@ -119,18 +121,35 @@ References:
...
     model::M
     p::P
+    st::ST
     re::RE
     basedist::D
     tspan::T
     args::A
     kwargs::K
+
+    function FFJORD(model::Lux.Chain,tspan,args...;p=nothing,st=nothing,basedist=nothing,kwargs...)
```
ChrisRackauckas, line 134: relax

src/ffjord.jl
ChrisRackauckas: This isn't correct. `m` is the model, not the result of the model. Flux `re(p)(x)` is equal to Lux `model(x,p,st)`. You should dispatch on whether `Flux.destructure` is done by making one dispatch for `re::Lux.AbstractExplicitLayer`, i.e. it should be:

```julia
function ffjord(u, p, t, model::AbstractExplicitLayer, e, st;
                regularize=false, monte_carlo=true)

function ffjord(u, p, t, re, e;
                regularize=false, monte_carlo=true)
```

for the Lux vs Flux cases.

Abhishek-1Bhatt:
The FFJORD example is close to working now, there's just this solve causing error

https://github.com/Abhishek-1Bhatt/DiffEqFlux.jl/blob/db27816708f3c62186d62e003ce154dba8f58541/src/ffjord.jl#L280
 
with a huge stacktrace. The error goes like this(total 89 stacks)
```julia
ERROR: Mutating arrays is not supported -- called setindex!(::Vector{Float32}, _...)
Stacktrace:  
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] (::Zygote.var"#437#438"{Vector{Float32}})(#unused#::Nothing)
    @ Zygote C:\Users\user\.julia\packages\Zygote\DkIUK\src\lib\array.jl:71
  [3] (::Zygote.var"#2337#back#439"{Zygote.var"#437#438"{Vector{Float32}}})(Δ::Nothing)
    @ Zygote C:\Users\user\.julia\packages\ZygoteRules\AIbCs\src\adjoint.jl:67
  [4] (::Zygote.var"#208#209"{Tuple{Tuple{Nothing, Nothing}, Tuple{Nothing}}, Zygote.var"#2337#back#439"{Zygote.var"#437#438"{Vector{Float32}}}})(Δ::Nothing)
    @ Zygote C:\Users\user\.julia\packages\Zygote\DkIUK\src\lib\lib.jl:207
  [5] (::Zygote.var"#1750#back#210"{Zygote.var"#208#209"{Tuple{Tuple{Nothing, Nothing}, Tuple{Nothing}}, Zygote.var"#2337#back#439"{Zygote.var"#437#438"{Vector{Float32}}}}})(Δ::Nothing)
    @ Zygote C:\Users\user\.julia\packages\ZygoteRules\AIbCs\src\adjoint.jl:67
  [6] macro expansion
    @ C:\Users\user\.julia\packages\ComponentArrays\lQw2S\src\array_interface.jl:129 [inlined]
  [7] Pullback
    @ C:\Users\user\.julia\packages\ComponentArrays\lQw2S\src\array_interface.jl:126 [inlined]
  [8] Pullback
    @ C:\Users\user\.julia\packages\ComponentArrays\lQw2S\src\namedtuple_interface.jl:17 [inlined]
  [9] Pullback
    @ C:\Users\user\.julia\packages\ComponentArrays\lQw2S\src\compat\chainrulescore.jl:4 [inlined]
 [10] (::typeof(∂(λ)))(Δ::Nothing)
    @ Zygote C:\Users\user\.julia\packages\Zygote\DkIUK\src\compiler\interface2.jl:0
 [11] Pullback
    @ C:\Users\user\.julia\packages\Zygote\DkIUK\src\compiler\chainrules.jl:205 [inlined]
 [12] (::typeof(∂(λ)))(Δ::Nothing)
    @ Zygote C:\Users\user\.julia\packages\Zygote\DkIUK\src\compiler\interface2.jl:0
```

ChrisRackauckas:
I assume @avik-pal hasn't tested nested reverse-mode AD with Lux yet. Try incorporating https://github.com/SciML/DiffEqFlux.jl/pull/614 into this. Even if forward-mode AD like that doesn't work, `SparseDiffTools.num_jacvec` should work, and there's absolutely no reason to nest reverse mode here anyways because that's just a slow way to compute something that is already inaccurate (stochastic approximation) anyways.

src/neural_de.jl
```diff
@@ -36,29 +37,37 @@ References:
 [1] Pontryagin, Lev Semenovich. Mathematical theory of optimal processes. CRC press, 1987.
 
 """
-struct NeuralODE{M,P,RE,T,A,K} <: NeuralDELayer
+struct NeuralODE{M,P,ST,RE,T,A,K} <: NeuralDELayer
     model::M
     p::P
+    st::ST
```
avik-pal, line 43: type of state is not guaranteed to be constant. Since this is a breaking change anyways, do we want to store the parameters and state in the struct of allow the user to handle that like in Lux?

ChrisRackauckas: Good point, I guess when constructing with a Lux chain it should return state and require calling the layers with state.

ChrisRackauckas:
The GPU test is fine on master: https://buildkite.com/julialang/diffeqflux-dot-jl/builds/829#01812ef5-81c3-4533-abca-bbd045a8f21e, looks like there might be a merge error.

Abhishek-1Bhatt:
Yes, now that we have imported Lux.jl, the uses of Dense and Chain have to be qualified. Let me make it Flux.Dense and Flux.Chain in the tests for now and if any tests need changes we can handle it in #715

ChrisRackauckas:
@avik-pal is there a reason for not supporting LTS? https://github.com/avik-pal/Lux.jl/blob/main/Project.toml#L37

ChrisRackauckas:
Let's merge anyways for now, but that should get fixed upstream.

avik-pal:
> @avik-pal is there a reason for not supporting LTS? https://github.com/avik-pal/Lux.jl/blob/main/Project.toml#L37

JET.jl need 1.7 for the tests

ChrisRackauckas:
Could the JET tests just be disabled on v1.6?

PR #724, opened 2022-06-14
Fixed NeuralDSDE

Abhishek-1Bhatt:
Fixed dereferencing of parameters in NeuralDSDE here for https://github.com/SciML/DiffEqSensitivity.jl/pull/623 to work , Lux compatible constructors for all layers have been added in SciML/DiffEqFlux.jl/pull/722

### DIFF
--- a/src/neural_de.jl
+++ b/src/neural_de.jl
@@ -159,7 +159,7 @@ struct NeuralDSDE{M,P,RE,M2,RE2,T,A,K} <: NeuralDELayer
         re2 = nothing
         new{typeof(model1),typeof(p),typeof(re1),typeof(model2),typeof(re2),
             typeof(tspan),typeof(args),typeof(kwargs)}(p,
-            length(p1),model1,re1,model2,re2,tspan,args,kwargs)
+            Int(1),model1,re1,model2,re2,tspan,args,kwargs)
     end
 end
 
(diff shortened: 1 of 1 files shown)


Abhishek-1Bhatt:
It expects `p` to be `[p1,p2]` . Doing it like `p = ComponentArray(p1, p2)` would lead to issues while dereferincing because
the user may name p1 and p2 anything but the functor expects it to be dereferenced by p.p1 and p.p2 . The new approach allows us to index p like a normal vector of containing two `ComponentArray`s.

src/neural_de.jl
```diff
@@ -179,13 +179,13 @@ function (n::NeuralDSDE{M})(x,p=n.p) where {M<:FastChain}
     solve(prob,n.args...;sensealg=TrackerAdjoint(),n.kwargs...)
 end
 
-function (n::NeuralDSDE{M})(x,p1,p2,st1,st2) where {M<:Lux.AbstractExplicitLayer}
+function (n::NeuralDSDE{M})(x,p,st1,st2) where {M<:Lux.AbstractExplicitLayer}
```
prbzrg, line 182: julia can unpack it
```suggestion
function (n::NeuralDSDE{M})(x,(p1, p2),st1,st2) where {M<:Lux.AbstractExplicitLayer}
```

Abhishek-1Bhatt: Actually we need them to be packed in one p so that we can pass it to the SDEProblem 
https://github.com/Abhishek-1Bhatt/DiffEqFlux.jl/blob/b3b7ae7208f03918bdaaf67b39b6775515dcc6e0/src/neural_de.jl#L193

which will then call our functions where we will dereference individual p1 and p2 from p

prbzrg: That's why I added parentheses it works with one `p` and Julia will unpack it like
```julia
p1, p2 = p
```

prbzrg: Oh, sorry. Now, I understand!

Issue #727, opened 2022-06-18
DiffEqFlux Layers don't satisfy Lux API


avik-pal:
The DiffEqFlux Layers need to satisfy https://lux.csail.mit.edu/dev/api/core/#Lux.AbstractExplicitLayer else the parameters/states returned from `Lux.setup` be incorrect. As pointed out in slack

```
julia> ps, st = Lux.setup(rng, Chain(node,Dense(2=>3)))
((layer_1 = NamedTuple(), layer_2 = (weight = Float32[0.11987843 -0.1679378; 0.36991563 0.41324985; 0.73272866 0.7062624], bias = Float32[0.0; 0.0; 0.0;;])), (layer_1 = NamedTuple(), layer_2 = NamedTuple()))
```

`ps.layer_1` should not be an empty NamedTuple

https://lux.csail.mit.edu/dev/manual/interface/ -- is the most recent manual for the interface

MilkshakeForReal:
Should be easy
```julia
initialparameters(rng::AbstractRNG, node::NeuralODE) = initialparameters(rng, node.model) 
initialstates(rng::AbstractRNG, node::NeuralODE) = initialstates(rng, node.model)
parameterlength(node::NeuralODE) = parameterlength(node.model)
statelength(node::NeuralODE) = statelength(node.model)
```
To make `setup` work, we need to add 
```julia
function setup(rng::AbstractRNG, node::NeuralODE)
    return (initialparameters(rng, node), initialstates(rng, node))
end
```

ChrisRackauckas:
Are you supposed to overload `setup`? I assume that should just follow from the interface.

avik-pal:
You just need to define

```julia
initialparameters(rng::AbstractRNG, node::NeuralODE) = initialparameters(rng, node.model) 
initialstates(rng::AbstractRNG, node::NeuralODE) = initialstates(rng, node.model)
```

ChrisRackauckas:
We should put an abstract type on all of the AbstractNeuralDE types and then overload from there.

MilkshakeForReal:
> You just need to define
> 
> ```julia
> initialparameters(rng::AbstractRNG, node::NeuralODE) = initialparameters(rng, node.model) 
> initialstates(rng::AbstractRNG, node::NeuralODE) = initialstates(rng, node.model)
> ```

For it to work yes. Would it be nicer if the number of parameters could be printed automatically?

MilkshakeForReal:
[...]

PR #735, opened 2022-06-21
Neural SDE tutorial

ChrisRackauckas:
Based off of the old one. Might be easiest to just make this Flux based?

### DIFF
--- a/docs/src/examples/neural_sde.md
+++ b/docs/src/examples/neural_sde.md
@@ -33,9 +33,8 @@ First let's build training data from the same example as the neural ODE:
 
 ```@example nsde
 using Plots, Statistics
-using Lux, Optimization, OptimizationFlux, DiffEqFlux, StochasticDiffEq, SciMLBase.EnsembleAnalysis, Random
+using Flux, Optimization, OptimizationFlux, DiffEqFlux, StochasticDiffEq, SciMLBase.EnsembleAnalysis
 
-rng = Random.default_rng()
 u0 = Float32[2.; 0.]
 datasize = 30
 tspan = (0.0f0, 1.0f0)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
@Abhishek-1Bhatt you're finishing this one? I think this is the last one to finish DiffEqFlux (for now).

docs/src/examples/neural_sde.md
```diff
@@ -0,0 +1,206 @@
...
+                       Lux.Dense(50, 2))
+p1, st1 = Lux.setup(rng, drift_dudt)
+
+diffusion_dudt = Lux.Chain(Lux.Dense(2, 2))
+p2, st2 = Lux.setup(rng, diffusion_dudt)
+
+p1 = Lux.ComponentArray(p1)
+p2 = Lux.ComponentArray(p2)
+#Component Arrays doesn't provide a name to the first ComponentVector, only subsequent ones get a name for dereferencing
+p = [p1, p2]
```
avik-pal, line 87: I will write a manual section later tonight in Lux itself, but this not quite the right interface. The "correct" thing to do would be 
```suggestion
p = Lux.ComponentArray((p1 = p1, p2 = p2))
```
Then index the parameters with `.p1` and `.p2`. This makes it consistent with the Lux ContainerLayer Interface.

The same thing holds for the states

Abhishek-1Bhatt: Would it be alright to make the user follow a strict naming scheme of p1 and p2, just a small concern.

avik-pal: So the user should not be responsible for constructing this in general. Lux.setup on NeuralDSDE should return the parameters in this format

avik-pal: The convention I follow in Lux is to document what the parameter and state names are http://lux.csail.mit.edu/stable/devdocs/style_guide/#documentation-for-layers, and construct the structure in `Lux.setup`. After that, if the user messes up then nothing can be done but redirect them to the documentation

Abhishek-1Bhatt: Okay, lets finalize this for all the Lux constructors in DiffEqFlux then

ChrisRackauckas: Yes, the user shouldn't construct this in general. This is just a start from a copy paste of broken code so don't take it as a correct way to do things for this. As mentioned in https://github.com/SciML/DiffEqFlux.jl/issues/727, these should be an `AbstractExplicitLayer`, which means they should do things exactly like `Dense`. They should have one state, take in a state, and return a state. They should take in a neural network definition and give you back a state from `setup`. Basically, it should act exactly like Dense does, and be able to perfectly swap in without any other code changes, and if not it's wrong. The only thing that should be different is the constructor for the layer.

Abhishek-1Bhatt:
We can merge #738 for now as discussed in https://github.com/SciML/DiffEqFlux.jl/pull/736#issuecomment-1163634256 and switch back to Lux with minor edits as soon as the issues are resolved.

PR #741, opened 2022-06-23
Fix latex typos in Hamiltonian Neural Network Tutorial

Algopaul:
The latex commands `\dot(x)` and `\ddot(x)` are replaced with `\dot x` and `\ddot x`, as parentheses should not be used here, as they lead to an incorrect rendering.


ChrisRackauckas:
Thanks!

PR #747, opened 2022-07-05
don't create a plot if not displaying it

chriselrod:


chriselrod:
> Is that slice a major deal? The plotting allocates will convert it anyways?

It could be hoisted out.
It isn't a big deal performance-wise, mostly just aesthetically speaking.

I don't like seeing fruit in reach, even if small.

PR #750, opened 2022-07-11
Lux compatible layers

Abhishek-1Bhatt:


src/neural_de.jl
```diff
@@ -183,33 +182,21 @@ function (n::NeuralDSDE{M})(x,p=n.p) where {M<:FastChain}
...
     function g(u,p,t;st=st2)
-      u_, st = n.model2(u,p.p2,st)
+      u_, st = n.model2(u,p.model2,st)
       return u_
     end
     
     ff = SDEFunction{false}(dudt_,g,tgrad=basic_tgrad)
     prob = SDEProblem{false}(ff,g,x,n.tspan,p)
-    return solve(prob,n.args...;sensealg=InterpolatingAdjoint(),n.kwargs...), (state1 = st1, state2 = st2)
+    return solve(prob,n.args...;sensealg=ReverseDiffAdjoint(),n.kwargs...), (model1 = st1, model2 = st2)
```
ChrisRackauckas, line 199: Why change the adjoint here?

src/neural_de.jl
```diff
@@ -1,6 +1,8 @@
-abstract type NeuralDELayer <: Lux.AbstractExplicitLayer end
+abstract type NeuralDELayer <: Lux.AbstractExplicitContainerLayer{(:model,)} end
+abstract type NeuralSDELayer <: Lux.AbstractExplicitContainerLayer{(:model1,:model2,)} end
```
avik-pal, line 2: I think we can call them `drift` and `diffusion` for more clarity

src/neural_de.jl
```diff
@@ -1,6 +1,8 @@
...
 basic_tgrad(u,p,t) = zero(u)
 Flux.trainable(m::NeuralDELayer) = (m.p,)
+Flux.trainable(m::NeuralSDELayer) = (m.p,)
```
avik-pal, line 5: I believe this API is deprecated and now you need to return a named tuple

src/neural_de.jl
```diff
@@ -183,33 +182,21 @@ function (n::NeuralDSDE{M})(x,p=n.p) where {M<:FastChain}
...
-    st2 = Lux.initialstates(rng, n.model2)
-    return (state1 = st1, state2 = st2)
-end
-
 function (n::NeuralDSDE{M})(x,p,st) where {M<:Lux.AbstractExplicitLayer}
-    st1 = st.state1
-    st2 = st.state2
+    st1 = st.model1
+    st2 = st.model2
     function dudt_(u,p,t;st=st1)
```
avik-pal, line 188: Why st=st1? This will make st a local variable and the one in the actual function scope will not be updated.

src/neural_de.jl
```diff
@@ -296,32 +283,21 @@ function (n::NeuralSDE{P,M})(x,p=n.p) where {P,M<:FastChain}
...
     function g(u,p,t;st=st2)
-        u_, st = n.model2(u,p.p2,st)
+        u_, st = n.model2(u,p.model2,st)
         return u_
     end
 
     ff = SDEFunction{false}(dudt_,g,tgrad=basic_tgrad)
     prob = SDEProblem{false}(ff,g,x,n.tspan,p,noise_rate_prototype=zeros(Float32,length(x),n.nbrown))
-    solve(prob,n.args...;sensealg=InterpolatingAdjoint(),n.kwargs...), (state1 = st1, state2 = st2)
+    return solve(prob,n.args...;sensealg=ReverseDiffAdjoint(),n.kwargs...), (model1 = st1, model2 = st2)
```
avik-pal, line 300: ReverseDiff (+ Lux) is not well tested. I would not make it the default

avik-pal (changes requested):

ChrisRackauckas:
What's the status of this?

Abhishek-1Bhatt:
NeuralDSDE and NeuralSDE with Lux work now using EM or LambaEM with BacksolveAdjoint. But the AD 'vjp's are throwing some errors. It works only with numerical `vjp` right now. I think we'll need some changes to SciMLSensitivity to make vjp compatible?

ChrisRackauckas:
How is it erroring? Use EulerHeun so it's in Stratanovich form: that's simpler for training.

test/neural_de_lux.jl
```diff
@@ -234,7 +234,7 @@ grads = Zygote.gradient(()->sum(sode(xs)),Flux.params(xs,sode))
 @test ! iszero(grads[sode.p])
 @test ! iszero(grads[sode.p][end])
 
-sode = NeuralDSDE(luxdudt,luxdudt2,(0.0f0,.1f0),LambaEulerHeun(),saveat=0.0:0.01:0.1)
+sode = NeuralDSDE(luxdudt,luxdudt2,(0.0f0,.1f0),EulerHeun(),saveat=0.0:0.01:0.1,dt=0.1)
```
ChrisRackauckas, line 237: why does a fixed `dt` make sense?

Abhishek-1Bhatt: The adaptive scheme was throwing the dtmin warning that says aborting. This one was working flawlessly for vector input. This too is failing for the matrix with some mul! error I'm looking into it rn

Abhishek-1Bhatt: Oh, I should've made that a dt of 0.01

avik-pal: +1 on the `mul!` error. There seems to be some dimension mismatch problem. I worked around it in my research codebases using:

```julia
# This is a weird type piracy but is needed to make the NeuralDSDE to work
function LinearAlgebra.mul!(A::AbstractVector{T}, B::AbstractMatrix{T},
                            C::AbstractMatrix{T}, b1::Bool, b2::Bool) where {T}
  LinearAlgebra.mul!(reshape(A, :, 1), B, reshape(C, :, 1), b1, b2)
  return A
end
```

ChrisRackauckas: But why is it there?

Abhishek-1Bhatt: Also [this](https://github.com/SciML/StochasticDiffEq.jl/blob/3d56e689a14d759e5ad0ac7eddc2cd0bc4b0bab3/src/perform_step/low_order.jl#L92) `mul!` is hitting the dispatch defined at [LeastSquaresOptim](https://github.com/matthieugomez/LeastSquaresOptim.jl/blob/main/src/solver/iterative_lsmr.jl) dispatch for some reason

Abhishek-1Bhatt:
```julia
Error During Test at /home/runner/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:25
  Got exception outside of a @test
  LoadError: MethodError: no method matching mul!(::Vector{Float32}, ::Matrix{Float32}, ::Matrix{Float64}, ::Bool, ::Bool)
  Closest candidates are:
    mul!(::Any, ::LeastSquaresOptim.PreconditionedMatrix{TA, Tp, Tx}, ::Any, ::Number, ::Number) where {TA, Tp, Tx} at ~/.julia/packages/LeastSquaresOptim/B1WKJ/src/solver/iterative_lsmr.jl:30
    mul!(::Any, ::LeastSquaresOptim.DampenedMatrix, ::Any, ::Number, ::Number) where T at ~/.julia/packages/LeastSquaresOptim/B1WKJ/src/solver/iterative_lsmr.jl:87
    mul!(::Any, ::LeastSquaresOptim.MyAdjoint{LeastSquaresOptim.PreconditionedMatrix{TA, Tp, Tx}}, ::Any, ::Number, ::Number) where {TA, Tp, Tx} at ~/.julia/packages/LeastSquaresOptim/B1WKJ/src/solver/iterative_lsmr.jl:36
```

ChrisRackauckas: That's not hitting the LeastSquaresOptim dispatches, it's just showing that those exist. The issue is that the last term should be `Matrix{Float32}`. Where is the non-diagonal noise matrix defined? I would assume it would be this code.

ChrisRackauckas: Is it coming from here https://github.com/SciML/SciMLSensitivity.jl/blob/master/src/concrete_solve.jl#L733 ? Are the parameters Float64?

Abhishek-1Bhatt: It being `Matrix{Float64}` is a part of the problem, but also two matrices are being passed to mul! to give a vector. I just converted `W.dW` directly to run:

```julia
ERROR: MethodError: no method matching mul!(::Vector{Float32}, ::Matrix{Float32}, ::Matrix{Float32}, ::Bool, ::Bool)
Closest candidates are:
  mul!(::Any, !Matched::LeastSquaresOptim.MyAdjoint{LeastSquaresOptim.PreconditionedMatrix{TA, Tp, Tx}}, ::Any, ::Number, ::Number) where {TA, Tp, Tx} at C:\Users\abhishek_bhatt\.julia\packages\LeastSquaresOptim\B1WKJ\src\solver\iterative_lsmr.jl:36
  mul!(::Any, !Matched::LeastSquaresOptim.MyAdjoint{LeastSquaresOptim.DampenedMatrix{TA, Tx}}, ::Any, ::Number, ::Number) where {TA, Tx} at C:\Users\abhishek_bhatt\.julia\packages\LeastSquaresOptim\B1WKJ\src\solver\iterative_lsmr.jl:95
  mul!(::Any, !Matched::LeastSquaresOptim.PreconditionedMatrix{TA, Tp, Tx}, ::Any, ::Number, ::Number) where {TA, Tp, Tx} at C:\Users\abhishek_bhatt\.julia\packages\LeastSquaresOptim\B1WKJ\src\solver\iterative_lsmr.jl:30
```

ChrisRackauckas: Change it to `LinearAlgebra.mul!`? Is that package shadowing `mul!`?

ChrisRackauckas: Why is that package in the image at all BTW? What even uses LeastSquaresOptim?

Abhishek-1Bhatt: That's what I was wondering, none of the packages involved depend on it

avik-pal: > The issue is that the last term should be Matrix{Float32}.

I don't think that is the problem. It is the `Matrix x Matrix` = `Vector` that is causing the error.

```julia
using LinearAlgebra

A = zeros(2); B = ones(2, 2); C = ones(2, 1);

LinearAlgebra.mul!(A, B, C)
# ERROR: MethodError: no method matching mul!(::Vector{Float64}, ::Matrix{Float64}, ::Matrix{Float64}, ::Bool, ::Bool)
# Closest candidates are:
#   mul!(::StridedVector{T}, ::StridedVecOrMat{T}, ::StridedVector{T}, ::Number, ::Number) where T<:Union{Float32, Float64, ComplexF32, ComplexF64} at /mnt/julia/juliaup/julia-1.8.2+0.x64/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:65
#   mul!(::StridedMatrix{T}, ::StridedVecOrMat{T}, ::StridedVecOrMat{T}, ::Number, ::Number) where T<:Union{Float32, Float64, ComplexF32, ComplexF64} at /mnt/julia/juliaup/julia-1.8.2+0.x64/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:159
#   mul!(::StridedVecOrMat, ::SymTridiagonal, ::StridedVecOrMat, ::Number, ::Number) at /mnt/julia/juliaup/julia-1.8.2+0.x64/share/julia/stdlib/v1.8/LinearAlgebra/src/tridiag.jl:217
#   ...
# Stacktrace:
#  [1] mul!(C::Vector{Float64}, A::Matrix{Float64}, B::Matrix{Float64})
#   @ LinearAlgebra /mnt/julia/juliaup/julia-1.8.2+0.x64/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:276
#  [2] top-level scope
#   @ REPL[3]:1
```

src/neural_de.jl
ChrisRackauckas, line 5:
```suggestion
```

ChrisRackauckas:
@avik-pal is that definition correct? https://github.com/SciML/DiffEqFlux.jl/actions/runs/3855851152/jobs/6571364733#step:6:777

avik-pal:
I had never stored parameters in the layers so never tested `@functor`. The fallback definition there is to allow recursively traverse container layers.

For Flux compat, I think we should directly define the `Optimisers.trainable` though that will not allow `Flux.gpu` to work as expected.

ChrisRackauckas:
But it seems it shouldn't dispatch on the second argument if it's going to work with `@functor`?

avik-pal:
https://github.com/avik-pal/Lux.jl/pull/229 should fix it. I will tag a release once the CI passes

ChrisRackauckas:
@Abhishek-1Bhatt https://github.com/SciML/DiffEqFlux.jl/actions/runs/3855851152/jobs/6581195485 this failure looks real, but looks like it just needs a functor overload.

Abhishek-1Bhatt:
The [error](https://github.com/SciML/DiffEqFlux.jl/actions/runs/3855851152/jobs/6581195485#step:6:2173) message, seems to be referring to `model` as in  `NeuralDELayer <: Lux.AbstractExplicitContainerLayer{(:model,)}` as `HamiltonianNeuralDE` is also subtyped to `NeuralDELayer` [here](https://github.com/SciML/DiffEqFlux.jl/blob/bf1ef54ca3181d7b012f6c0bdfc371b393e41328/src/hnn.jl#L92) , maybe doing something like [this](https://github.com/Abhishek-1Bhatt/DiffEqFlux.jl/blob/efb435f34dc655c957bf0c5bed2bae8fe8cbda80/src/neural_de.jl#L652) could help?

test/runtests.jl
```diff
@@ -11,7 +11,8 @@ if GROUP == "All" || GROUP == "DiffEqFlux" || GROUP == "Layers"
 end
 
 if GROUP == "All" || GROUP == "DiffEqFlux" || GROUP == "BasicNeuralDE"
-    @safetestset "Neural DE Tests" begin include("neural_de.jl") end
+    @safetestset "Neural DE Tests with Lux" begin include("neural_de_lux.jl") end
+    #@safetestset "Neural DE Tests" begin include("neural_de.jl") end
```
ChrisRackauckas, line 15:
```suggestion
    @safetestset "Neural DE Tests" begin include("neural_de.jl") end
```

Issue #745, opened 2022-07-02
Warning: `ignore(f)` is deprecated, use `ChainRulesCore.ignore_derivatives(f)` instead.


prbzrg:
I got these warnings:
```julia
┌ Warning: `ignore(f)` is deprecated, use `ChainRulesCore.ignore_derivatives(f)` instead.
│   caller = forward_ffjord(n::DiffEqFlux.FFJORD{Chain{Tuple{Dense{typeof(tanh), Matrix{Float64}, Vector{Float64}}, Dense{typeof(tanh), Matrix{Float64}, Vector{Float64}}, Dense{typeof(tanh), Matrix{Float64}, Vector{Float64}}}}, Vector{Float64}, Nothing, Optimisers.Restructure{Chain{Tuple{Dense{typeof(tanh), Matrix{Float64}, Vector{Float64}}, Dense{typeof(tanh), Matrix{Float64}, Vector{Float64}}, Dense{typeof(tanh), Matrix{Float64}, Vector{Float64}}}}, NamedTuple{(:layers,), Tuple{Tuple{NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}, NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}, NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}}}}}, Distributions.DiagNormal, Tuple{Float64, Float64}, Tuple{OrdinaryDiffEq.Tsit5{typeof(OrdinaryDiffEq.trivial_limiter!), typeof(OrdinaryDiffEq.trivial_limiter!), Static.True}}, Base.Pairs{Symbol, SciMLSensitivity.InterpolatingAdjoint{0, true, Val{:central}, SciMLSensitivity.ZygoteVJP}, Tuple{Symbol}, NamedTuple{(:sensealg,), Tuple{SciMLSensitivity.InterpolatingAdjoint{0, true, Val{:central}, SciMLSensitivity.ZygoteVJP}}}}}, x::Matrix{Float64}, p::Vector{Float64}, e::Matrix{Float64}; regularize::Bool, monte_carlo::Bool) at ffjord.jl:285
└ @ DiffEqFlux C:\Users\Hossein Pourbozorg\.julia\packages\DiffEqFlux\7N0N5\src\ffjord.jl:285
[...]

PR #751, opened 2022-07-16
update Zygote.@ignore to ChainRulesCore.@ignore_derivatives

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqFlux.jl/issues/745

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "1.51.2"
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 Cassette = "7057c7e9-c182-5462-911a-8362d720325c"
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
@avik-pal it looks like there's a regression because of the type piracy:

```julia
function ChainRulesCore.rrule(::typeof(convert), T::DataType, x::AbstractMatrix)
    y = convert(T, x)
    function convert_pullback(dy)
        if dy isa NoTangent || dy isa ZeroTangent
            dx = dy
        else
            dx = convert(typeof(x), dy)
        end
        return NoTangent(), NoTangent(), dx
    end
    return y, convert_pullback
end
```

Is this essential for Lux? Is there a reason this isn't upstreamed?

PR #753, opened 2022-07-23
Dimension Mover Layer

### LINKED ISSUE #111: Suggestion:  Change dimension order on NeuralODE to match convention of ML batch

prbzrg:
Closes #111

### DIFF
--- a/docs/src/layers/NeuralDELayers.md
+++ b/docs/src/layers/NeuralDELayers.md
@@ -15,3 +15,10 @@ NeuralDAE
 NeuralODEMM
 AugmentedNDELayer
 ```
+
+# Helper Layer Functions
+
+```@docs
+DimMover
+FluxBatchOrder
+```
--- a/src/DiffEqFlux.jl
+++ b/src/DiffEqFlux.jl
@@ -62,6 +62,7 @@ export HamiltonianNN
 export ChebyshevBasis, SinBasis, CosBasis, FourierBasis, LegendreBasis, PolynomialBasis
 export FastDense, StaticDense, FastChain, initial_params
 export FFJORDDistribution
+export DimMover, FluxBatchOrder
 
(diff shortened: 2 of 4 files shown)


ChrisRackauckas:
Add to the docs?

Issue #62, opened 2019-07-17
Register


ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

ChrisRackauckas:
@JuliaRegistrator register()

[...]

PR #772, opened 2022-10-10
Update workflows

prbzrg:
I just updated some workflows:

- used `julia-actions/cache` instead of `actions/cache`
- skipped intermediate builds
- added a formatter workflow
- added a workflow to use `julia-actions/RegisterAction`

And I also added some tests for quality assurance.
I know `CompatHelper` file got an update recently, but IMO, it's better to use `julia-actions/setup-julia` instead of OS binary, so we can control the Julia version and also be sure it's official binary of Julia.


ChrisRackauckas:
Seems like this is getting failures

prbzrg:
I couldn't figure out the problem, probably related to deprecation of `DeterministicCNF`.
Can we delete it and only keep [the `@deprecate` line](https://github.com/SciML/DiffEqFlux.jl/blob/df3c5f620d461b1c818ccb9d0d0ade1b18dda9ff/src/ffjord.jl#L64) to not break anything?

ChrisRackauckas:
Try it

Could we do the quality check as a separate PR script that we can easily copy-paste everywhere?

.github/workflows/Register.yml
```diff
@@ -0,0 +1,14 @@
+name: Register Package
+on:
```
ChrisRackauckas, line 2: what's this?

.github/workflows/Formatter.yml
```diff
@@ -0,0 +1,37 @@
+name: Formatter
```
ChrisRackauckas, line 1: wasn't there already a format yml?

prbzrg: The workflow in #740 is to check if a PR complies to the style guide, but this workflow is for continuous checking the main branch and creates a PR automatically if needed.

ChrisRackauckas: We settled on not doing that so that the git blame would more readily point back to the correct PR

.github/workflows/Register.yml
prbzrg: With this workflow, it registers a new version of the package just by specifying which part of the version to increase (major, minor, or patch).

More info:
https://github.com/julia-actions/RegisterAction

ChrisRackauckas: But we have JuliaRegistrator setup everywhere. Any reason to not be using JuliaRegistrator?

prbzrg: It works based on JuliaRegistrator. if we have it, there is no need to [manually increase the version](https://github.com/SciML/DiffEqFlux.jl/commit/092af896745e338d29c929b557e8728d4ef55327), or [trigger a release by commenting in a PR](https://github.com/SciML/DiffEqFlux.jl/issues/62).

ChrisRackauckas: Ehh let's stick to JuliaRegistrator for now. It's quite a bit of an overhaul to train everyone on a new system if it doesn't give a tangible improvement. Also, updating all of the repos will be a nightmare.

ChrisRackauckas:
Cool, do you have the mass PR ready?

prbzrg:
This PR is ready to review.

ChrisRackauckas:
I think it looks good. Will you run the mass action today?

prbzrg:
If you mean `MassInstallAction`, no, I can't do it.

ChrisRackauckas:
Okay then we can wait a bit. Actions should be uniform across the repos.

PR #775, opened 2022-10-23
canonize docs

ArnoStrouwen:


Vaibhavdixit02:
Created a new release of OptimizationOptimJL which should fix the failing docs issues here I think, let's re-run and confirm before merging

ChrisRackauckas:
Let's do that separately, since this won't break anything. Open a test master PR.

PR #780, opened 2022-11-21
Add quality check

prbzrg:
I added a quality check using `Aqua`.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,11 +1,24 @@
-using DiffEqFlux, SafeTestsets, Test
+using DiffEqFlux, Aqua, SafeTestsets, Test
 
 const GROUP = get(ENV, "GROUP", "All")
 const is_APPVEYOR = (Sys.iswindows() && haskey(ENV,"APPVEYOR"))
 const is_CI = haskey(ENV,"CI")
 
 @time begin
-if GROUP == "All" || GROUP == "DiffEqFlux" || GROUP == "Layers"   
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -1,11 +1,24 @@
...
 @time begin
-if GROUP == "All" || GROUP == "DiffEqFlux" || GROUP == "Layers"   
+if GROUP == "All" || GROUP == "DiffEqFlux" || GROUP == "Quality"
+    Aqua.test_ambiguities(DiffEqFlux)
+    Aqua.test_unbound_args(DiffEqFlux)
+    Aqua.test_undefined_exports(DiffEqFlux)
+    Aqua.test_project_extras(DiffEqFlux)
+    Aqua.test_stale_deps(DiffEqFlux)
+    Aqua.test_deps_compat(DiffEqFlux)
+    Aqua.test_project_toml_formatting(DiffEqFlux)
```
ChrisRackauckas, line 15: can we make this into some standard script that we can paste to every repo without having to modify anything but the YAML?

prbzrg: I think we can read the package name from `Project.toml` and `eval` it. If we could use [`Aqua.test_all`](https://juliatesting.github.io/Aqua.jl/dev/#Aqua.test_all-Tuple{Module}), it would be cleaner, but it looks like injection of `Core` makes warnings for other imported packages and [fails the test](https://github.com/SciML/DiffEqFlux.jl/actions/runs/3223804188/jobs/5274255464).

Issue #788, opened 2022-12-30
Wrong order in collocations


axla-io:
https://github.com/SciML/DiffEqFlux.jl/blob/ccc3dcd0ccb7c18a176fc30dc9cb964acf98d50b/src/collocation.jl#L158

This line should be 
`du, u = collocate_data(reshape(data, 1, :),tpoints,tpoints_sample,interp,args...)`

ChrisRackauckas:
definitely.

ChrisRackauckas:
How was that passing tests...

axla-io:
Yeah that's pretty surprising

PR #789, opened 2022-12-30
Fix flipped collocation

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqFlux.jl/issues/788


ChrisRackauckas:
The test worked because it was a linear function lol https://github.com/SciML/DiffEqFlux.jl/blob/ChrisRackauckas-patch-1/test/collocation_regression.jl#L4

Issue #744, opened 2022-06-25
Move AD Rules to `SciMLSensitivity`


prbzrg:
https://github.com/SciML/DiffEqFlux.jl/blob/e32422d86bea1cdda8797d0b909cb3e207738bd5/src/DiffEqFlux.jl#L24-L40

Related to https://github.com/SciML/SciMLSensitivity.jl/issues/582

ChrisRackauckas:
Those rules should just be removed or upstreamed. @oxinabox @DhairyaLGandhi are any of those required these days?

prbzrg:
What's update on this?

ChrisRackauckas:
Tridiagonal rules moved with https://github.com/JuliaDiff/ChainRules.jl/pull/758

ChrisRackauckas:
Oh, these all were removed awhile ago.

PR #715, opened 2022-06-04
Remove deprecations for breaking release

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqFlux.jl/issues/707

PR #794, opened 2023-01-17
Deprecate FastChain and sciml_train for v2.0

ChrisRackauckas:
Superscedes https://github.com/SciML/DiffEqFlux.jl/pull/715


test/multiple_shoot.jl
```diff
@@ -23,16 +24,17 @@ prob_trueode = ODEProblem(trueODEfunc, u0, tspan)
 ode_data = Array(solve(prob_trueode, Tsit5(), saveat = tsteps))
 
 # Define the Neural Network
-nn = FastChain((x, p) -> x.^3,
-                FastDense(2, 16, tanh),
-                FastDense(16, 2))
-p_init = initial_params(nn)
+nn = Lux.Chain(ActivationFunction(x -> x.^3),
```
ChrisRackauckas, line 27:
```suggestion
nn = Lux.Chain(x -> x.^3,
```

test/neural_dae.jl
```diff
@@ -41,3 +41,29 @@ p = p .+ rand(3) .* p
 optfunc = Optimization.OptimizationFunction((x, p) -> loss(x), Optimization.AutoZygote())
 optprob = Optimization.OptimizationProblem(optfunc, p)
 res = Optimization.solve(optprob, BFGS(initial_stepnorm = 0.0001))
+
+# Same stuff with Lux
+rng = Random.default_rng()
+dudt2 = Lux.Chain(ActivationFunction(x -> x.^3),Lux.Dense(6,50,tanh),Lux.Dense(50,2))
```
ChrisRackauckas, line 47:
```suggestion
dudt2 = Lux.Chain(x -> x.^3,Lux.Dense(6,50,tanh),Lux.Dense(50,2))
```

test/multiple_shoot.jl
```diff
@@ -23,16 +24,17 @@ prob_trueode = ODEProblem(trueODEfunc, u0, tspan)
...
 # Define the Neural Network
-nn = FastChain((x, p) -> x.^3,
-                FastDense(2, 16, tanh),
-                FastDense(16, 2))
-p_init = initial_params(nn)
+nn = Lux.Chain(x -> x.^3,
+                Lux.Dense(2, 16, tanh),
+                Lux.Dense(16, 2))
+p_init, st = Lux.setup(rng, nn)
+p_init = Lux.ComponentArray(p_init)
```
ChrisRackauckas, line 31: @Abhishek-1Bhatt did you test Lux with multiple shooting?

test/newton_neural_ode.jl
```diff
@@ -23,24 +22,23 @@ nODE = NeuralODE(NN, tspan, ROCK4(), reltol=1f-4, saveat=[tspan[end]])
...
 l1 = loss_function(nODE.p)
+optf = Optimization.OptimizationFunction((x,p)->loss_function(x), Optimization.AutoZygote())
+optprob = Optimization.OptimizationProblem(optf, nODE.p)
 
-res = DiffEqFlux.sciml_train(loss_function, nODE.p, NewtonTrustRegion(), Optimization.AutoZygote(), maxiters=100, cb=cb) #ensure backwards compatibility of `cb`
+res = Optimization.solve(optprob, NewtonTrustRegion(), maxiters=100, callback=cb)
 @test loss_function(res.minimizer) < l1
-res = DiffEqFlux.sciml_train(loss_function, nODE.p, Optim.KrylovTrustRegion(), Optimization.AutoZygote(), maxiters = 100, callback=cb)
+res = Optimization.solve(optprob, KrylovTrustRegion(), maxiters=100, callback=cb)
 @test loss_function(res.minimizer) < l1
```
ChrisRackauckas, line 31: We should probably also test second order Lux.

Project.toml
```diff
@@ -20,8 +18,6 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 LoggingExtras = "e6f89c97-d47a-5376-807f-9c37f3926c36"
 Lux = "b2108857-7c20-44ae-9111-449ecde12c47"
-NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
-Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 Optimization = "7f7a1694-90dd-40f0-9382-eb1efda571ba"
 OptimizationFlux = "253f991c-a7b2-45f8-8852-8b9a9df78a86"
```
prbzrg, line 22: Shouldn't this be removed?

prbzrg: `Optimization`, `OptimizationFlux`, and `OptimizationOptimJL` are not imported.

ChrisRackauckas: yes

prbzrg:
I think `Flux` optimizers get converted to `Optimisers` ones. So it's better to replace `OptimizationFlux` with `OptimizationOptimisers` in tests. (I'm not sure)

ChrisRackauckas:
> I think Flux optimizers get converted to Optimisers ones. So it's better to replace OptimizationFlux with OptimizationOptimisers in tests. (I'm not sure)

Yes that is true, but I don't want to mix too many changes at the same time. Since that's separate from DiffEqFlux itself, I'd like to hold off on that. But definitely we want to get rid of Flux.jl and Optim.jl optimizer usage.

prbzrg:
IMO, for a breaking release, it's better to resolve #744 first. and remove `ZygoteRules` too.

ChrisRackauckas:
Yes that should be in the release too. I think it can just be removed and the tests will still pass? I guess just give it a try? If Newton and FFJORD pass then we're good.

src/DiffEqFlux.jl
```diff
@@ -1,23 +1,13 @@
...
-@reexport using Zygote
-
-# deprecate
-
-import OptimizationFlux
-import NNlib
-import Lux
-using Requires
-using Cassette
 @reexport using Flux
```
prbzrg, line 10: Now that we support `Lux`, reexporting `Flux` could make qualification errors in codes that only have `Lux`.
Error like:
```julia
WARNING: both Lux and Flux export "Chain"; uses of it in module Main must be qualified
```

prbzrg: #795

ChrisRackauckas: Yeah we should remove the reexport. Add that to the list of things.

test/mnist_gpu.jl
```diff
@@ -57,7 +57,7 @@ m = Flux.Chain(down,
     DiffEqArray_to_Array,
     fc) |> gpu
```
Abhishek-1Bhatt, line 58:
```suggestion
    fc) |> Flux.gpu

```

Abhishek-1Bhatt: Also here

https://github.com/SciML/DiffEqFlux.jl/blob/e019eb799e80cf0ad9e4d55766013c4d757c1e63/test/mnist_gpu.jl#L50

Abhishek-1Bhatt: I think its happening because we have `using Lux` here instead of the `import Lux` in master . But we can resolve these namespace issues by just depending on `LuxCore` since the layer types are all we need from Lux

ChrisRackauckas: Yeah in DiffEqFlux we should probably only use LuxCore.

ChrisRackauckas:
I'm trying to just get this to mergeable now. The plan will be to hold breaking on master for a month (which we normally don't do, but we can make an exception for this big of a deprecation). We should do all of the things like remove the Zygote rule overloads (make sure they are upstreamed), setup LuxCore, update docs to not use OptimizationFlux, etc. and get this 2.0 fully ready. But this keeps being a runaway train of a PR so I want to get something that works, and then continue with improvements that can be done in parallel.

ChrisRackauckas:
🎉 master now in v2.0 mode, slam it with the breaking changes you want.

PR #804, opened 2023-01-21
Switch to LuxCore

Abhishek-1Bhatt:


src/ffjord.jl
```diff
@@ -35,25 +34,21 @@ References:
...
-            p = _p
-        end
+    function FFJORD(model::LuxCore.AbstractExplicitLayer,tspan,args...;p=nothing,st=nothing,basedist=nothing,kwargs...)
+        re = nothing
+        p = nothing
         if isnothing(basedist)
-            size_input = size(p.layer_1.weight, 2)
-            type_input = eltype(p.layer_1.weight)
+            size_input = model.layers.layer_1.in_dims
+            type_input = eltype(tspan)
```
Abhishek-1Bhatt, line 51: Would this be alright?

ChrisRackauckas: I don't see what the issue would be.

PR #820, opened 2023-03-08
Drop the `Flux` and have it only as an extension

prbzrg:
- [X] remove Flux parts and use `Lux.transform`
- [ ] clean any Flux related patterns (re, ...)
- [ ] convert it to an extension

I guess now `DiffEqFlux` is actually `DiffEqFLux`. 😏 (F stands for "for")


ChrisRackauckas:
> I guess now DiffEqFlux is actually DiffEqFLux. 😏 (F stands for "for")

Beautiful. Poetry. Master.

prbzrg:
This isn't working, I will make a new PR.

PR #827, opened 2023-05-18
missing deps on node example

anandijain:


ChrisRackauckas:
It looks like this wasn't branched off of the real master?

anandijain:
ah actually the stable example is broken but dev works. what is the way to get stable docs working?

anandijain:
nevermind dev was broken too

PR #853, opened 2023-09-02
fix: kernel functions

sathvikbhagavan:
`abs(t) > 0` will give all points except 0. This means K(t) will result in 0 for all points except 0. That is incorrect.


ChrisRackauckas:
Is there a nice test for this?

sathvikbhagavan:
What kind of tests should I add? Unit tests for all the kernels or regression tests similar to https://github.com/SciML/DiffEqFlux.jl/blob/master/test/collocation_regression.jl?

I see only `SigmoidKernel` was used in https://github.com/SciML/DiffEqFlux.jl/blob/master/test/collocation_regression.jl, so the bug was never caught.

ChrisRackauckas:
> What kind of tests should I add? Unit tests for all the kernels or regression tests similar to https://github.com/SciML/DiffEqFlux.jl/blob/master/test/collocation_regression.jl?

That would be good.

PR #858, opened 2023-09-24
Documenter 1.0 upgrade.

ArnoStrouwen:
Also adds formatter workflow in anticipation of formatter being fixed.
Formatting has thus not yet been applied.


ChrisRackauckas:
@ArnoStrouwen can you rebase this? And how do we find that lux link?

ChrisRackauckas:
we can just point to the lux docs

ArnoStrouwen:
The SciML multi documenter is not able to find the Lux docs:
https://docs.sciml.ai/Lux/
https://github.com/SciML/SciMLDocs/blob/main/docs/make_aggregate.jl#L134C32-L134C63
Lux has some extra things going on with Documenter, namely Vitepress:
https://github.com/LuxDL/Lux.jl/blob/main/docs/make.jl
Is this the reason that it fails?
@mortenpi could you please give your opinion?

ChrisRackauckas:
Just link to the lux website and remove from here.

ArnoStrouwen:
Is this a known stochastic failure? Was not there the last time it ran, and only a link changed.
https://buildkite.com/julialang/diffeqflux-dot-jl/builds/1353#018b0509-9480-40bb-9631-66f15bdd5220/896-1547

ArnoStrouwen:
Should be good to go except for the weird failure.

ChrisRackauckas:
Complete!

mortenpi:
> Lux has some extra things going on with Documenter, namely Vitepress:

Yeah, it looks like Lux is nowadays [committing (Vitepress) Markdown files to `gh-pages`](https://github.com/LuxDL/Lux.jl/tree/gh-pages/v0.5.5) and then using that to deploy the full site in the `gh-pages` workflow. This means that we don't really have access to the generated HTML, and so can't really include that in a MultiDocumenter build.

Note that the SciML aggregate builds are not broken per se, it's just they're pushing Markdown files to S3: https://docs.sciml.ai/Lux/stable/manual/weight_initializers.md

So I feel that Lux should probably be removed from the aggregate site build?

PR #887, opened 2023-12-17
docs: run all the example blocks and clean up a few things

sathvikbhagavan:
Noticed a few examples were not run in the doc build.

## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API

PR #890, opened 2023-12-23
docs: run all the example blocks and clean up a few things

avik-pal:
Adding a few fixes on top of #887


ChrisRackauckas:
https://buildkite.com/julialang/diffeqflux-dot-jl/builds/1495#018c98fc-74ea-4dd9-8c80-8a166b2d8ff1/1462-4793

Lol the plots are too big on one example and that's it.

PR #898, opened 2024-01-12
docs: remove ude and mention neural ode

sathvikbhagavan:
As DiffEqFlux is separated from SciMLSensitivity stuff, DiffEqFlux does not house anything related to UDEs.

I also noticed the links to the papers in https://docs.sciml.ai/DiffEqFlux/dev/#Pre-Built-Architectures to be not what I was expecting:
1. Neural SDEs point to Zygote's paper - https://arxiv.org/abs/1907.07587
2. Neural DAEs point to UDE paper - https://arxiv.org/abs/2001.04385
3. Neural DDEs also point to UDE paper - https://arxiv.org/abs/2001.04385
4. CNFs point to Neural ODE paper - https://arxiv.org/abs/1806.07366

I am not sure if that's intentional.

Also, the citation section consists of UDE paper, is there a DiffEqFlux paper to replace it?

Also, we don't have tutorials to NeuralDAEs, NeuralDDEs by looking at the list of current functionality and current tutorials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 DiffEq(For)Lux.jl (aka DiffEqFlux.jl) fuses the world of differential equations with machine learning
 by helping users put diffeq solvers into neural networks. This package utilizes
 [DifferentialEquations.jl](https://docs.sciml.ai/DiffEqDocs/stable/), and [Lux.jl](https://lux.csail.mit.edu/)  as its building blocks to support research in
-[Scientific Machine Learning](https://www.stochasticlifestyle.com/the-essential-tools-of-scientific-machine-learning-scientific-ml/), specifically neural differential equations and universal differential equations, to add physical information into traditional machine learning.
+[Scientific Machine Learning](https://www.stochasticlifestyle.com/the-essential-tools-of-scientific-machine-learning-scientific-ml/), specifically neural differential equations to add physical information into traditional machine learning.
 
 > [!NOTE]
 > We maintain backwards compatibility with [Flux.jl](https://docs.sciml.ai/Flux/stable/) via [Lux.transform](https://lux.csail.mit.edu/stable/api/Lux/flux_to_lux#Lux.transform)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
> I also noticed the links to the papers in https://docs.sciml.ai/DiffEqFlux/dev/#Pre-Built-Architectures to be not what I was expecting:

Why would those not be what's expected? Those are the first sources.

README.md
```diff
@@ -27,25 +27,20 @@ the documentation, which contains the unreleased features.
...
-
-As such, it is the first package to support and demonstrate:
-
-  - Stiff and non-stiff universal ordinary differential equations (universal ODEs)
-  - Universal stochastic differential equations (universal SDEs)
-  - Universal delay differential equations (universal DDEs)
-  - Universal partial differential equations (universal PDEs)
-  - Universal jump stochastic differential equations (universal jump diffusions)
-  - Hybrid universal differential equations (universal DEs with event handling)
+DiffEqFlux.jl is for neural ordinary differential equations.
```
ChrisRackauckas, line 30:
```suggestion
DiffEqFlux.jl is for implicit layer machine learning.
```

ChrisRackauckas:
> Also, the citation section consists of UDE paper, is there a DiffEqFlux paper to replace it?

No, it's basically the same paper in the end. There was a blog post the evolved into the UDE paper. This code then branched off of what is now SciMLSensitivity as just standard architectures.

ChrisRackauckas:
> Also, we don't have tutorials to NeuralDAEs, NeuralDDEs by looking at the list of current functionality and current tutorials.

That's because as implicit layer ML methods, I don't think they are very good and there are clear technical reasons why. So they work and train but I don't have good examples because I don't think there's an example that is better done with another architecture.

sathvikbhagavan:
>Why would those not be what's expected? Those are the first sources.

Oh, I thought there might be specialized papers for NeuralDAEs and NeuralDDEs .  I also missed the part of NeuralSDEs in the end of the Zygote's paper. That clarifies it.

I think this PR can be merged then.

PR #930, opened 2024-06-13
Spline Tests should be fixed now

avik-pal:


ChrisRackauckas:
Those test failures don't look so good 😓 . Is there an easy way to turn off the multithreading with the whole ReTest thing?

PR #937, opened 2024-08-28
refactor: use layers from `Boltz`

### LINKED ISSUE #929: Restructuring some of the Layers
**Question❓**

What do we feel about moving some of the non-diffeq specific layers to `Boltz.jl` and re-exporting them from here?

I was working on implementing KANs and we might be able to reuse the Spline Layers from here, but DiffEqFlux has too heavy deps (not complaining here, it needs to). Boltz.jl on the other hand is quite lightweight, only pulling in stdlibs and Lux (and some other smallish packages).

avik-pal:
fixes #929 

needs https://github.com/LuxDL/Boltz.jl/pull/54


avik-pal:
Let's merge this once the CI passes. Doing the Lux 1.0 update here would bloat the PR too much

PR #950, opened 2024-09-24
Bump compats and update tutorials for Optimization v4

Vaibhavdixit02:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API


Project.toml
```diff
@@ -34,7 +34,7 @@ Boltz = "1"
 ChainRulesCore = "1"
 ComponentArrays = "0.15.17"
 ConcreteStructs = "0.2"
-DataInterpolations = "5, 6"
```
avik-pal, line 37: don't remove 5

Vaibhavdixit02: That causes conflict too, with Symbolics

Vaibhavdixit02: Just saw your comment on the compat PR in Boltz, I'll find a workaround

ChrisRackauckas: Why would 5 be needed?

Vaibhavdixit02:
This is stalled and will likely lead to more people complaining about the Optimization change can we get a decision for it?

Project.toml
ChrisRackauckas: Fixed

ChrisRackauckas:
@avik-pal update Boltz?

avik-pal:
Some tests still fail for DataInterpolations, let me dev and patch those shouldn't be hard.

avik-pal:
https://github.com/JuliaRegistries/General/pull/116656

docs/src/examples/multiple_shooting.md
```diff
@@ -62,14 +65,13 @@ end
...
+function callback(state, l; doplot = true, prob_node = prob_node)
     display(l)
     global iter
     iter += 1
     if doplot && iter % 1 == 0
         # plot the original data
         plt = scatter(tsteps, ode_data[1, :]; label = "Data")
-
         # plot the different predictions for individual shoot
         plot_multiple_shoot(plt, preds, group_size)
```
ChrisRackauckas, line 76: I don't see how this is going to work if we don't have the predictions?

docs/src/examples/neural_ode.md
```diff
@@ -40,15 +40,16 @@ end
...
 end
 
 # Do not plot by default for the documentation
 # Users should change doplot=true to see the plots callbacks
-callback = function (p, l, pred; doplot = false)
+function callback(state, l; doplot = false)
     println(l)
     # plot current prediction against data
     if doplot
+        pred = predict_neuralode(state.u)
```
ChrisRackauckas, line 52: This is not a viable solution, it doubles the cost. We should show a better standard way to do this.

Vaibhavdixit02: It was the same before the gradient and loss were evaluated separately so I figured it's a fair tradeoff, the current alternative is to use a global variable.

Down the line I plan to improve OptimizationState so it can be hacked for these things, I am not sure how it should be yet but should be possible

ChrisRackauckas: We should show a current best practice that does not involve recomputing since that can be quite a regression.

ChrisRackauckas:
What's the state of this?

docs/src/examples/hamiltonian_nn.md
```diff
@@ -23,37 +23,33 @@ p_t = reshape(cos.(2π_32 * t), 1, :)
...
 ps, st = Lux.setup(Xoshiro(0), hnn)
 ps_c = ps |> ComponentArray
 
 opt = OptimizationOptimisers.Adam(0.01f0)
 
-function loss_function(ps, data, target)
+function loss_function(ps, databatch)
+    data, target = databatch
     pred, st_ = hnn(data, ps, st)
     return mean(abs2, pred .- target), pred
```
ChrisRackauckas, line 41:
```suggestion
    return mean(abs2, pred .- target)
```

Vaibhavdixit02:
Yay! 🎉

PR #962, opened 2025-01-24
Included definition of v in this example

macquarrielucas:
Previously v was left undefined in this example. I added the system of ODEs you can from this second order differential equation so that readers know that v is. Perhaps this is obvious from physics perspective but not necessarily from other practices.

## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API

### DIFF
--- a/docs/src/examples/tensor_layer.md
+++ b/docs/src/examples/tensor_layer.md
@@ -9,6 +9,13 @@ Let's consider the anharmonic oscillator described by the ODE
 ẍ = - kx - αx³ - βẋ -γẋ³.
 ```
 
+We first transform this second order differential equation into a system of first order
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
yeah oops, everyone has a different background so explicit is always better. Thanks!

PR #1025, opened 2026-06-07
Re-enable downgrade CI (group matrix) with corrected [compat] lower bounds

ChrisRackauckas-Claude:
Re-enables the disabled downgrade workflow as a **group matrix** mirroring Tests.yml (Layers, BasicNeuralDE, AdvancedNeuralDE, Newton, QA) — an unset GROUP would run no tests. `allow-reresolve: true` (strict `false` is infeasible here: latest SciMLBase 3.18 pulls latest NonlinearSolve → the upstream LogExpFunctions-1.0.1 ForwardDiff≤0.10.22 cap). Floors raised to the consistent set (SciMLBase 3.18, SciMLSensitivity 7.112, OptimizationOptimJL 0.4.14, DelayDiffEq 6.0.3, Lux 1.29, ForwardDiff 1.3.3, OrdinaryDiffEq 7.0.0, …).

**Verification:** ✅ Layers, BasicNeuralDE, Newton, QA groups PASS locally on Julia 1.10 at these floors. ⚠️ AdvancedNeuralDE (CNF/FFJORD) is heavy and was not confirmed locally (run interrupted under load) — CI verifies it.

> Ignore until reviewed by @ChrisRackauckas.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ChrisRackauckas-Claude:
Pushed a follow-up commit removing the now-invalid `allow-reresolve:` input from `Downgrade.yml`. The reusable `SciML/.github` `downgrade.yml` workflow is being changed to always use `allow_reresolve: false` and to drop the `allow-reresolve` `workflow_call` input entirely (see SciML/.github PR). Once that .github PR merges and `@v1` is retagged, any caller still passing `allow-reresolve:` would error with an invalid-input failure, so this line must go. Behavior is unchanged (it was already strict/false here).

PR #1049, opened 2026-08-14
Drop older SciML major versions from compat

ChrisRackauckas-Claude:
**Please ignore this PR until it has been reviewed by @ChrisRackauckas.**

## What changed

Pure `[compat]` edits that **drop older major versions of SciML packages**. For each SciML dependency, only the current major remains (existing lower bound on that major is kept). No code, tests, or package version were changed.

- `Project.toml` [extras] `DataInterpolations`: `8.9, 9.0` → `9.0` (drop 8.9; current SciML major is 9)

## Why

Supporting multiple SciML majors keeps untested resolver windows. The current major is the one in the General registry; older majors are dropped even when that current major is recent.

## Verification

Current majors come from JuliaRegistries/General `Versions.toml`. Not verified here: the full test suite, QA, or a live `julia-downgrade-compat` run.


ChrisRackauckas-Claude:
Closing this draft: the proposed floor changes did not produce a passing Downgrade lane or regular test evidence, so it should not be kept as an unverified compatibility fix.

